PR #15, opened 2019-07-22
Minor changes and additions

VoOoLoX:


Craftplacer (approved):
The commands.json is not needed and probably won't because we don't have any plans to port the vanilla command system and its commands.

Ignoring that, everything seems fine.

VoOoLoX:
Great I was wondering if I should include commands.json or not, guess I'll remove it.
I think it would be nice to have a tool to extract blocks and items files from the actual jar as a separate project maybe or at least guide in the README because it took me some time to figure it out. Anyway I might look into it later.

Naamloos:
@Tides updates on review?

Craftplacer:
> @Tides updates on review?

bump

Tides (approved):
Sorry for the late response but all changes seem fine to me 😃

PR #19, opened 2020-08-23
Added SerializationMethodBuilder.

Seb-stian:
I added SerializationMethodBuilder, which uses DynamicMethods instead of Reflection to increase performance. I also added UnitTests for deserialization. So far, only IncomingChatMessage packet deserialization is implemented.

Usage is the same:
`var message = await PacketSerializer.FastDeserializeAsync<IncomingChatMessage>(packet.data);`

### DIFF
--- a/Obsidian.Tests/Deserialization.cs
+++ b/Obsidian.Tests/Deserialization.cs
@@ -1,0 +1,94 @@
+﻿using Obsidian.Net;
+using Obsidian.Net.Packets.Play;
+using Obsidian.Serializer;
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Obsidian.Tests
+{
+    public class Deserialization
+    {
+        [Fact]
+        public void Handshake()
+        {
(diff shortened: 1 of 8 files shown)


Obsidian.Tests/Deserialization.cs
```diff
@@ -0,0 +1,94 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Obsidian.Tests
+{
+    public class Deserialization
+    {
+        [Fact]
+        public void Handshake()
```
Craftplacer: It's not a good idea to include failing tests at the moment, the missing tests should be TODO comments, so we can take care about implementing them later

Craftplacer (changes requested):
Looks good, just we should have tests always pass in our latest code.

Craftplacer (approved):

Obsidian.Tests/Deserialization.cs
Seb-stian: I commented out all tests that weren't implemented.

PR #20, opened 2020-08-24
Implemented more packets and packet tests.

Seb-stian:
I implemented the rest of deserialization, including tests.

### DIFF
--- a/Obsidian/Serializer/Enums/DataType.cs
+++ b/Obsidian/Serializer/Enums/DataType.cs
@@ -20,7 +20,13 @@
         EntityMetadata,
         Slot,
         NbtTag,
+        /// <summary>
+        /// Position serialized as an Int64.
+        /// </summary>
         Position,
+        /// <summary>
+        /// Position serialized as three double values.
+        /// </summary>
         AbsolutePosition,
         Transform,
         SoundPosition,
--- a/Obsidian/Serializer/Enums/DataType.cs
+++ b/Obsidian/Serializer/Enums/DataType.cs
@@ -2,7 +2,7 @@
 {
     public enum DataType
     {
-        Auto = 0,
+        Auto,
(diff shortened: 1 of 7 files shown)


Obsidian/Serializer/Enums/DataType.cs
```diff
@@ -2,7 +2,7 @@
 {
     public enum DataType
     {
-        Auto = 0,
+        Auto,
```
Craftplacer, line 5: Was the null redundant?

Obsidian/Serializer/Enums/DataType.cs
```diff
@@ -21,10 +21,14 @@ public enum DataType
         Slot,
         NbtTag,
         Position,
+        AbsolutePosition,
```
Craftplacer, line 30: Mind telling what's the difference between Position and AbsolutePosition?

Craftplacer (approved):

Obsidian/Serializer/Enums/DataType.cs
Tides: Absolute position uses the Position class but se/de-serializes by reading/writing the x, y and z separately while Position would usually mean you're getting/writing the values from/to a long

Craftplacer: Hmm, weird name choice.
Might be worth documenting this encoding on the enum value?

Tides: Sounds good, lets do that

Tides (approved):

Obsidian/Serializer/Enums/DataType.cs
Seb-stian: The 0 was redundant, yes.

PR #43, opened 2020-11-02
Fiddling with MainCommandModule & console stuffs

Czompi:
**Modified commands**
- Added command usage argument to CommandInfo
- Added #region for every command
- Added command info for every command
- Reworked /help command
- Added /gamemode command
- Added /stop command, that will stop the server (like in vanilla-based servers)
- Added SendCommandUsage(string) function to easily display command usage (without duplicating this code for every scenario).


Naamloos (approved):
Seems fine, though usage _could_ be constructed using the parameters given in a command. oh right, overloads lol

PR #47, opened 2020-11-04
Packet changes 2.0

Tides:
This removes the abstract class Packet and makes packet use a new Interface called IPacket ^ ^


Obsidian/Util/PacketHandler.cs
```diff
@@ -67,528 +37,69 @@ public static async Task<Packet> ReadCompressedPacketAsync(MinecraftStream strea
...
-                                break;
-
-                            case BlockFace.East:
-                                location.X += 1;
-                                break;
-
-                            default:
-                                break;
-                        }
+        public static void RegisterHandlers()
```
Naamloos, line 44: an idea: we could make an attribute for packets with an id and define default behavior for IPacket.Id to return the attribute's value. But keep it like this for now, I'll experiment with it once it hits stable.

Naamloos (approved):

PR #49, opened 2020-11-06
Add IncomingChatMessage event.

oliver4888:


Seb-stian (approved):
Maybe we should make it implement ICancellable, so that it can override the default behavior. We can add that later though and other than that I approve. Thanks for the PR!

Tides (approved):
Looks good, I'll merge it and work on making it cancellable so people can override basic chat ~

PR #59, opened 2020-11-11
Small permission system update

roxxel:


Naamloos (changes requested):
Perhaps more useful would be a HasAllPermissions method?

If you're not willing to add it to this PR just say so and it's ready to merge.

roxxel:
> Perhaps more useful would be a HasAllPermissions method?
> 
> If you're not willing to add it to this PR just say so and it's ready to merge.

Okay, i'll add it, also what about having both HasAnyPermission and HasAllPermissions methods, and attributes as well?

TheBunnies:
Both **HasAnyPermission** and **HasAllPermissions** are useful, separate attributes are required ofc. And I was actually thinking if it's better to move **HasAllPermissions** and **HasAnyPermissions** to the extension methods.

Obsidian.API/CommandFramework/Attributes/RequirePermissionAttribute.cs
```diff
@@ -6,21 +6,22 @@ namespace Obsidian.Commands
...
 
         public override async Task<bool> RunChecksAsync(ObsidianContext ctx)
         {
             if (this.op && ctx.Player.IsOperator)
                 return true;
 
-            if (this.permission != "")
-                return await ctx.Player.HasPermission(permission.ToLower());
+            if (this.permissions != null)
+                return await ctx.Player.HasAnyPermission(permissions);
```
Craftplacer: Wouldn't we expect that all permissions match? Maybe an enum for telling Obsidian if you want to match All, Any or None permissions.

e.g.
`[RequirePermissions(PermissionCheckType.All, "obsidian.shitpost", "obsidian.anime")]`

I won't require to add this, just document it somewhere using xmldocs.

Craftplacer (approved):

Obsidian.API/CommandFramework/Attributes/RequirePermissionAttribute.cs
Naamloos: Craft is right, the attribute should check ALL perms by default, not ANY.

TheBunnies: Yep, that's actually a good idea.

Obsidian.API/CommandFramework/ArgumentParsers/StringArgumentParser.cs
```diff
@@ -1,7 +1,4 @@
-﻿using Obsidian.CommandFramework.Entities;
```
Naamloos, line 1: It wasn't necessary to remove unused usings in the same Pull Request..

Obsidian.API/CommandFramework/Attributes/RequirePermissionAttribute.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace Obsidian.Commands
+{
+    public class RequirePermissionAttribute : BaseExecutionCheckAttribute
+    {
+        private string[] permissions;
+        private PermissionCheckType checkType;
+        private bool op;
+
+        public RequirePermissionAttribute(PermissionCheckType checkType = PermissionCheckType.All, bool op = false, params string[] permissions)
```
Naamloos, line 13: please default op to true

Obsidian/Net/Packets/Play/Client/EntityVelocity.cs
```diff
@@ -1,12 +1,12 @@
...
+﻿using Obsidian.API;
 using Obsidian.Entities;
+using Obsidian.Serializer.Attributes;
+using Obsidian.Serializer.Enums;
 using System.Threading.Tasks;
 
 namespace Obsidian.Net.Packets.Play.Client
 {
-    public class EntityVelocity :  IPacket
+    public class EntityVelocity : IPacket
```
Naamloos, line 9: Can you please next time you make a PR _not_ do this stuff? At least do it in a separate PR because this is hell to review :p

Obsidian/WorldData/World.cs
```diff
@@ -94,6 +94,55 @@ public async Task UpdateClientChunksAsync(Client c, bool forcereload = false)
...
+                    if (chk is null)
+                    {
+                        chunksToGen.Add(new Position(cx, 0, cz));
+                    }
+                    else
+                    {
+                        await c.SendChunkAsync(chk);
+                    }
+                }
+            }
```
Naamloos, line 131: did you re-add removed methods???

Obsidian/WorldData/World.cs
```diff
@@ -376,6 +429,7 @@ internal void SetWorldSpawn()
                             if (by > 58 && (block.Type == Materials.GrassBlock || block.Type == Materials.Sand))
                             {
                                 Data.SpawnX = bx;
+                                Data.SpawnY = by + 1;
```
Naamloos, line 432: what??

Naamloos (changes requested):
Next time you open a PR, don't touch indentation and usings. it's hell to review and I'm not even sure we can merge this now without conflicts..

along with that, you seem to have readded methods that were removed from world.cs.

Please also change the op parameter of the Permission attribute to default to true.

PR #60, opened 2020-11-20
/Gamemode, bug fixes

Exeton:
Changes
  - Exceptions in HandlePlayPackets used to prevent future incoming packets from being handled
  - World#TotalLoadedEntites used to throw a null ref exception when not all regions are loaded (causing the server to crash if you joined before all regions were loaded.) Now if not all regions are loaded, you will not get sent all chunk data, but you can rejoin to fix this.
  - Exceptions in sendPacketBlock are now logged
  - PacketSerializer.SerializeAsync will now unlock MinecraftStream if there's an exception
  - Added conversion in MinecraftStream from enum to float. (This might be something that should be generalized to more dataTypes?)

### DIFF
--- a/Obsidian/Net/MinecraftStream.cs
+++ b/Obsidian/Net/MinecraftStream.cs
@@ -255,7 +255,14 @@ namespace Obsidian.Net
                     }
                 case DataType.Float:
                     {
-                        await this.WriteFloatAsync((float)value);
+						if (value is Enum)
+						{
+							Enum _enum = (Enum)value;
+                            value = Convert.ChangeType(value, _enum.GetTypeCode());
+							await this.WriteFloatAsync(Convert.ToSingle(value));
+						}
+						else
+							await this.WriteFloatAsync((float)value);
                         break;
                     }
                 case DataType.Double:
--- a/Obsidian/Net/MinecraftStream.cs
+++ b/Obsidian/Net/MinecraftStream.cs
@@ -255,14 +255,14 @@ namespace Obsidian.Net
                     }
                 case DataType.Float:
                     {
-						if (value is Enum)
-						{
-							Enum _enum = (Enum)value;
+                        if (value is Enum)
+                        {
+                            Enum _enum = (Enum)value;
                             value = Convert.ChangeType(value, _enum.GetTypeCode());
-							await this.WriteFloatAsync(Convert.ToSingle(value));
-						}
-						else
-							await this.WriteFloatAsync((float)value);
+                            await this.WriteFloatAsync(Convert.ToSingle(value));
+                        }
+                        else
+                            await this.WriteFloatAsync((float)value);
                         break;
                     }
                 case DataType.Double:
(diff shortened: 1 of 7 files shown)


Jonpro03 (changes requested):
There are mixed tabs/spaces in the PR. Can you convert the tabs to spaces pls?

Exeton:
Ok, I've converted tabs to spaces

Jonpro03 (approved):
LGTM. Thanks for the contribution. Pls standby for another reviewer.

Obsidian/Serializer/PacketSerializer.cs
```diff
@@ -21,48 +21,57 @@ public static class PacketSerializer
...
+                dataStream.Position = 0;
+                // await dataStream.DumpAsync(packet: packet);
 
-            stream.Lock.Release();
+                await dataStream.CopyToAsync(stream);
+            }
+            catch
+            {
+                throw;
+			}
```
Craftplacer: Formatting mistake

Obsidian/Net/MinecraftStream.cs
```diff
@@ -255,7 +255,14 @@ public async Task WriteAsync(DataType type, FieldAttribute attribute, object val
...
                     {
-                        await this.WriteFloatAsync((float)value);
+                        if (value is Enum)
+                        {
+                            Enum _enum = (Enum)value;
+                            value = Convert.ChangeType(value, _enum.GetTypeCode());
+                            await this.WriteFloatAsync(Convert.ToSingle(value));
+                        }
+                        else
+                            await this.WriteFloatAsync((float)value);
```
Craftplacer, line 265: I don't see why `float`'s should be ever mapped to `enum`s, is this the case anywhere? This doesn't have to removed though.

Craftplacer (changes requested):
Though Obsidian is still in development, having this detail of verbosity (printing exceptions of failed packets), is something that should not be enabled by default. Some server admins can't make use out of it (maybe when reporting bugs? but what about IO exceptions like inproper disconnect?).

Obsidian/Net/MinecraftStream.cs
Tides: ChangeGameState uses floats for its values ^ ^

Naamloos: I think this should be based on what type the enum uses under the hood (`enum thing : int`, or `enum thing : float`, etc)

Exeton: It can be an int. I checked to see if I could make it a float enum, but c# doesn't support it.

Craftplacer: If that's the case, leave it in.

blobcatcomfglare

Naamloos: ight, lgtm then

Naamloos:
Resolved the other requested change, but the formatting error should be fixed. Check the other requested changes :)

Obsidian/Serializer/PacketSerializer.cs
Exeton: I've made the exception logging off by default + fixed the invalid formatting in PacketSerializer.cs

Craftplacer (approved):

PR #76, opened 2021-01-10
Small visual improvement(s) and added element(s) to ETextAction

Czompi:
### Small visual improvement(s) and added element(s) to ETextAction
**Summary:**
- Modified the version that showing when you're using an unsupported version of Minecraft (`Obsidian v1_16_4` => `Obsidian 1.16.4`).
- Added `ETextAction.CopyToClipboard` because it was missing (I didn't check if others where missing because this was necessary for my *new* plugin.

**Added methods**
- `N/A`


Naamloos:
LGTM

Craftplacer (approved):
Hacky fix, we could consider using attributes for labeling versions correctly, but for now, this is fine.

PR #78, opened 2021-01-18
Command framework improvements

Naamloos:
## Description
Made some changes to the Commands Framework (and in addition, to the plugin manager and plugins).

## Most notable changes
- Commands can now be registered from a `PluginBase`.
- For commands in other classes, plugins can define dependencies. For this, I've introduced the `CommandDependencyBundle` class and the `RegisterCommandDependencies` method.
- Added the `RegisterSingleCommand` method to `PluginBase`. Requires a valid method with command attributes. At this moment it does not support references to it's defining class. I will update that later. The method might change. (*breaking change*)
- Command classes in plugin assemblies are now required to be registered with the `CommandRoot` attribute, so they can be found by Obsidian and be properly registered. (internally, this is not needed for the `MainCommandModule`, as we register that manually.)
- Add a fix that properly defines plugin commands to clients.
- The command framework itself has been moved back to the base server.
- Methods for registering commands have been removed from the IServer interface, this is now all automated via attributes. This is purely so it's easier to keep track of what plugin registered what command, making it possible to unregister commands later.
- Oh, right, you can unregister commands now. Useful for unloading plugins at runtime. Take note that this still requires a call to `Registry.RegisterCommands(Server);` to ensure the internal declare commands packet is updated.
- Added more love to the bottom of `PluginManager.cs`.
- `ObsidianContext` was renamed to `CommandContext`
- Command Framework is now located at `Commands/Framework` and now has the `Obsidian.Commands.Framework` namespace.

### Requesting review from other maintainers.

_note: I might have missed some things in this PR description._


Obsidian.API/_Types/CommandContext.cs
```diff
@@ -0,0 +1,39 @@
...
+    public class CommandContext
+    {
+        internal string Message;
+
+        public CommandContext(string message, IPlayer player, IServer server/*, IClient client*/)
+        {
+            this.Player = player;
+            this.Server = server;
+            this.Message = message;
+            //this.Client = client;
```
roxxel: Maybe would be better to remove this?

Naamloos: Yes, but not for this PR.

Naamloos:
Proposed changes:
- Use only `PluginBase` as a Dependency for command classes.
- Remove `CommandDependencyBundle`
- Add `PluginBase` to `CommandContext` instead of `Dependencies`
- Use `[Inject]` to inject `PluginBase` into command classes

***!!!Please do not merge yet!!!***

Naamloos:
Just went for injecting any dependency, and Plugins as an extension to that for Command Roots and subclasses.

PR #80, opened 2021-01-19
Hex colors support

roxxel:
#### Summary
- Added `HexColor` struct
- Added `color` property to ChatMessage

### DIFF
--- a/Obsidian.API/_Types/HexColor.cs
+++ b/Obsidian.API/_Types/HexColor.cs
@@ -1,0 +1,52 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Drawing;
+using System.Linq;
+using System.Text;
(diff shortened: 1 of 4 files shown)


Naamloos:
LGTM

Obsidian.API/_Types/HexColor.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        public HexColor(Color color)
+        { 
+            hexColor = $"#{color.R:X2}{color.G:X2}{color.B:X2}";
+        }
+
+        public HexColor(string hex)
+        {
+            hexColor = hex;
```
Craftplacer, line 27: Could be a point of failure (invalid value, also missing information which format it should be), but can be addressed in a single commit afterwards.

PR #84, opened 2021-01-29
Improve event system

Seb-stian:
- plugins can use methods that return `void` and `ValueTask` for minecraft events
- some implementation has been moved from the API
- added XML docs for some types in API.Events

I encountered some problems while writing `AsyncEvent`, so some of the code can be scuffed. Some methods could probably be simplified/removed?


Naamloos:
LGTM, might wanna update docs

Naamloos:
based that a PR made in january still has no conflicts

PR #86, opened 2021-02-19
Implemented SpawnExperienceOrb and SpawnPainting

roxxel:


Obsidian.API/_Types/Painting.cs
```diff
@@ -0,0 +1,58 @@
...
+namespace Obsidian.API
+{
+    public class Painting
+    {
+
+        public readonly short Id;
+        public readonly byte X;
+        public readonly byte Y;
+        public readonly byte Width;
+        public readonly byte Height;
```
Seb-stian: These could be properties.

_Since it's not a lot of data, `Painting` could even be a struct?_

Obsidian/WorldData/World.cs
```diff
@@ -483,5 +483,22 @@ internal bool TryAddEntity(Entity entity)
...
+                Position = position,
+                EntityId = 2 
+                // 2 = Experience Orb Source: https://minecraft.gamepedia.com/Java_Edition_data_values/Pre-flattening/Entity_IDs
+            });
+        }
+
+        public async Task SpawnPainting(Position position, Painting painting, PaintingDirection direction, Guid uuid = default)
+        {
+            if (uuid == Guid.Empty) uuid = Guid.NewGuid();
+            await this.Server.BroadcastPacketAsync(new SpawnPainting(9, uuid, painting.Id, position, direction));
```
Seb-stian: I wonder where does the magic `9` come from?

Obsidian/Net/Packets/Play/Clientbound/SpawnPainting.cs
```diff
@@ -0,0 +1,55 @@
...
+        public Guid UUID;
+
+        [Field(2), VarLength]
+        public int Motive;
+
+        [Field(3)]
+        public Position Position;
+
+        [Field(4), ActualType(typeof(byte))]
+        public PaintingDirection Direction;
```
Seb-stian: Also could be properties.

Seb-stian (approved):
Looks really good 👍

Obsidian.API/_Interfaces/IWorld.cs
```diff
@@ -7,5 +10,8 @@ public interface IWorld
 
         public long Time { get; }
         public Gamemode GameType { get; }
+
+        public Task SpawnExperienceOrb(PositionF position, short count);
```
Craftplacer: What does `count` mean? The variable name is unclear and misses XML docs for clarification.

Obsidian/WorldData/World.cs
Craftplacer: > I wonder where does the magic `9` come from?

The comment above seems related, https://minecraft.gamepedia.com/Java_Edition_data_values/Pre-flattening/Entity_IDs
says that 9 is the Entity ID for paintings.

Craftplacer (approved):

Obsidian.API/_Interfaces/IWorld.cs
Seb-stian: I'm pretty sure that it's the experience orb count. Maybe the method should be renamed to `SpawnExperienceOrbs`.

Naamloos: A rename of the method to more clearly define that it's multiple orbs would be nice, yes.

Obsidian.API/_Types/Painting.cs
Naamloos: +1 on that

Obsidian/WorldData/World.cs
Naamloos: If the entity ID for paintings is _always_ 9, it should be in `SpawnPainting.cs` and setting a different entity ID shouldn't even be an option in the constructor. I'd say either hard code it in the `SpawnPainting` constructor, on one of it's properties or look if there's any loaded data for entity IDs in the registry.

Naamloos (changes requested):
I recommend looking at the given feedback and make changes accordingly. It looks good overall, but some minor changes can be made. Once that is done, LGTM and merge ready.

but first, I'll mark this as a request for changes

Obsidian/WorldData/World.cs
Naamloos: Addition to that, if the entity id for paintings is static then it's kinda weird that mojang defines it in the packets. hmm

Obsidian.API/_Types/Painting.cs
```diff
@@ -6,14 +6,14 @@
...
-    public class Painting
+    public struct Painting
     {
 
-        public readonly short Id;
-        public readonly byte X;
-        public readonly byte Y;
-        public readonly byte Width;
-        public readonly byte Height;
+        public short Id { get; private set; }
```
Craftplacer: Properties can be set in the constructor no matter if the `set` is missing, I would suggest to remove it.

Craftplacer (approved):

Obsidian.API/_Types/Painting.cs
Seb-stian: And then even make the struct `readonly`.

Obsidian/Net/Packets/Play/Clientbound/SpawnExperienceOrb.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        [Field(1), Absolute]
+        public PositionF Position { get; private set; }
+
+        [Field(2)]
+        public short Count { get; private set; }
+
+        public int Id => 0x01;
+
+
```
Seb-stian, line 25: There should be an empty constructor (at least a private one). I don't think the project will even compile without it. (generators)

Seb-stian (changes requested):
Needs changes, see my SpawnExperienceOrb.cs comment.

Seb-stian (approved):

Seb-stian:
With approval from @Naamloos I would merge.

Naamloos (approved):
New changes look good to me, nice work 👌👍

PR #88, opened 2021-03-11
Rename and improve Position(F)

Seb-stian:
- Renamed `Position` and `PositionF` to `Vector` and `VectorF`, since their use is more generic than just as a position.
- Improved/added methods (`Cross`, `Dot`, `Lerp`...)
- `static Vector.ChunkClamped(int, int, int)` should be used instead of `Vector.ChunkClamp()` when possible


Seb-stian:
Relevant changes:
[Vector](https://github.com/ObsidianMC/Obsidian/pull/88/files#diff-c604d4dfbd50d1d9989fb00dda54c51ddb6130f94b8912f3377a61c5bb22cf1f)
[VectorF](https://github.com/ObsidianMC/Obsidian/pull/88/files#diff-409eb292316d02e1f41082feac31a7ff4c79419d40d5c691ae7d14d47cbb0e06)

Obsidian.API/Obsidian.API.xml
```diff
@@ -1040,163 +1040,300 @@
             A class that represents an angle from 0° to 360° degrees.
             </summary>
         </member>
-        <member name="T:Obsidian.API.Position">
+        <member name="T:Obsidian.API.Vector">
```
Naamloos, line 1043: I should add this file to the `.gitignore`, as it's a generated file.

Obsidian/Net/MinecraftStream.Writing.cs
```diff
@@ -396,15 +396,15 @@ public void WritePosition(Position value)
         }
 
         [WriteMethod, Absolute]
-        public void WriteAbsolutePosition(Position value)
+        public void WriteAbsolutePosition(API.Vector value)
```
Naamloos: Why is it API.Vector here, but just VectorF in other places?

Seb-stian: This was done by VS for some reason.

Naamloos (approved):
LGTM. My comments are just that, comments.

Craftplacer (approved):

Obsidian/Net/MinecraftStream.Writing.cs
Craftplacer: Explicit namespaces either mean a missing import or conflicting member names.

Craftplacer (approved):

PR #90, opened 2021-03-13
Voxel traversal

roxxel:


Obsidian.API/_Types/Traversal.cs
```diff
@@ -0,0 +1,102 @@
...
+
+            float tMaxX, tMaxY, tMaxZ, tDeltaX, tDeltaY, tDeltaZ;
+            VectorF voxel = new VectorF();
+
+            float x1 = StartPoint.X, y1 = StartPoint.Y, z1 = StartPoint.Z; // start point   
+            float x2 = EndPoint.X, y2 = EndPoint.Y, z2 = EndPoint.Z; // end point   
+
+
+            var dx = Math.Sign(x2 - x1); // x direction
+            if (dx != 0) tDeltaX = Math.Min(dx / (x2 - x1), 10000000.0f); else tDeltaX = 10000000.0f;
```
Seb-stian: `10000000.0f` seems to be a magic number. Could it be `float.MaxValue` or `float.PositiveInfinity`?

roxxel: Done

Obsidian/Entities/Entity.cs
```diff
@@ -28,22 +28,6 @@ public class Entity : IEquatable<Entity>, IEntity
         public Angle Pitch { get; set; }
 
         public Angle Yaw { get; set; }
-
```
Naamloos, line 31: Why exactly is this removed?

Obsidian.API/_Interfaces/IEntity.cs
```diff
@@ -8,7 +8,6 @@ public interface IEntity
         public VectorF Position { get; set; }
         public Angle Pitch { get; set; }
         public Angle Yaw { get; set; }
-        public VectorF LookDirection { get; }
```
Naamloos, line 11: same here

roxxel: Didn't work as expected

Obsidian/Entities/Entity.cs
roxxel: Just misclick

TheBunnies:
Sooo, I don't know if my opinion is matter here. But I have a few suggestions about this PR. First of all I don't understand why you're combining **var's** with explicit types, sometimes you write **List<VectorF> blocks = new()** and sometimes you write **var something = new()**. Second, why you use IEnumerable as a return object? IEnumerable makes a lot of sense in the method arguments but as a return type it adds more implicity in my opnion.

Naamloos:
@TheBunnies Thank you for your review! Anyone's opinion matters if what they say is relevant and truthful!

Issue #83, opened 2021-01-24
Limit command issuing


Craftplacer:
> Original discussion from https://github.com/ObsidianMC/Obsidian/issues/82#issuecomment-766437251

Commands will be able to be tagged with an `IssuerScopeAttribute` attribute, which takes an enum of following values:

- **`None`** Command is basically disabled *(probably remove this?)*
- **`Client`** Command sent by a player
- **`Console`** Command typed into the native server console
- **`RemoteConsole`** Command sent by remote console like [RCON](https://github.com/ObsidianMC/Obsidian/issues/82)
- **`Plugin`** Command executed by plugin
- **`Any`** Flags of all values above

By default, if a command doesn't specify a scope, it will be able to be executed by any issuer (player, console, plugin).

To not force our opinions on what should be executable or not, the config should be able to configure what scopes commands have.

PR #91, opened 2021-04-16
Implemented executing commands directly from console

roxxel:
### Summary
- Implemented executing commands directly from console
- Added ability to limit command issuing (#83)
- Small improvement of TPS counter


Obsidian/ConsoleClient.cs
```diff
@@ -0,0 +1,183 @@
...
+        public bool Sleeping { get => default; set => _ = value; }
+        public bool Sneaking { get => default; set => _ = value; }
+        public bool Sprinting { get => default; set => _ = value; }
+        public bool FlyingWithElytra { get => default; set => _ = value; }
+        public bool InHorseInventory { get => default; set => _ = value; }
+        public short AttackTime { get => default; set => _ = value; }
+        public short DeathTime { get => default; set => _ = value; }
+        public short HurtTime { get => default; set => _ = value; }
+        public short SleepTimer { get => default; set => _ = value; }
+        public short CurrentSlot { get => default; set => _ = value; }
```
Seb-stian: This looks very odd to me. Perhaps trying to access these properties (and methods) should throw `NotSupportedException` instead?

Obsidian/Program.cs
```diff
@@ -109,6 +109,87 @@ private static async Task Main()
...
+        {
+            new Thread(new ThreadStart(async () =>
+            {
+                try
+                {
+                    Server currentServer = Servers.First().Value;
+                    Thread.Sleep(2000);
+                    while (!shutdownPending)
+                    {
+                        if (currentServer == null && Servers.Count == 0 && !shutdownPending)
```
Seb-stian: `&& !shutdownPending` looks redundant. Also, wouldn't `||` make more sense than `&&`?

Obsidian/Program.cs
```diff
@@ -109,6 +109,87 @@ private static async Task Main()
             }
         }
 
+        private static void InitConsoleInput()
+        {
+            new Thread(new ThreadStart(async () =>
```
Seb-stian: Why is it not just a Task?

Obsidian/Program.cs
```diff
@@ -109,6 +109,87 @@ private static async Task Main()
...
+                            {
+                                Servers.Remove(currentServer.Id);
+                                currentServer = Servers.FirstOrDefault().Value;
+                            }
+                        }
+
+                    }
+                }
+                catch (IOException) { }
+                catch (NullReferenceException) { }
```
Seb-stian: This tells me, that something in there is going wrong 😄

Obsidian/Server.cs
```diff
@@ -290,6 +289,45 @@ public async Task StartServerAsync()
...
+                Logger.LogError($"No such command was found.");
+            }
+            catch (NoSuchParserException)
+            {
+                Logger.LogError($"The command you executed has a argument that has no matching parser.");
+            }
+            catch (InvalidCommandOverloadException)
+            {
+                Logger.LogError($"No such overload is available for this command.");
+            }
```
Seb-stian: I looked through `CommandHandler.ProcessComand` and whenever an exception is thrown, it always seems to set the `Exception.Message`, why aren't you using that?

Obsidian/Server.cs
```diff
@@ -559,7 +597,18 @@ private async Task ServerLoop()
 
             short itersPerSecond = 0;
             var stopWatch = Stopwatch.StartNew(); // for TPS measuring
-
+            _ = Task.Run(() =>
+            {
+                while (true)
+                {
+                    if (stopWatch.ElapsedMilliseconds >= 1000L)
```
Seb-stian: I'm not sure how good idea this is. My proposal would be to replace original code with something like
```cs
TPS = (short)(1.0 / stopwatch.Elapsed.TotalSeconds);
stopwatch.Restart();
```
which would remove the need of tracking how many iterations happened in a second (`itersPerSecond`) and would update the `TPS` value every tick, instead of every second.

Obsidian/Utilities/ConsoleIO.cs
```diff
@@ -0,0 +1,380 @@
...
+using System.Threading;
+
+namespace Obsidian.Utilities
+{
+
+    /// <summary>
+    /// Allows simultaneous console input and output without breaking user input
+    /// (Without having this annoying behaviour : User inp[Some Console output]ut)
+    /// Provide some fancy features such as formatted output, text pasting and tab-completion.
+    /// By ORelio - (c) 2012-2018 - Available under the CDDL-1.0 license
```
Seb-stian, line 14: I'm just wondering, is the license OK? Honestly, I'm not a big fan of how the whole class is implemented, so I might eventually rewrite it to something custom anyways.

Seb-stian, line 14: I'm just wondering, is the license OK? Honestly, I'm not a big fan of how the whole class is implemented, so I might eventually rewrite it to something custom anyways. Seems fine for now.

Obsidian/Utilities/Extensions.Colors.cs
```diff
@@ -24,7 +23,7 @@ public static void RenderColoredConsoleMessage(this string message)
                 // Print text with previous color
                 if (start != i)
                 {
-                    output.Write(message.AsSpan(start, i - start));
+                    ConsoleIO.Write(message.AsSpan(start, i - start).ToString());
```
Seb-stian, line 26: With this change, it might as well be `Substring`. (Also line 46)

Obsidian/Server.cs
roxxel: I copied this code from [here](https://github.com/ObsidianMC/Obsidian/blob/eab18e7946db785c49302046cfdbfed68d46a077/Obsidian/Server.cs#L303)

Obsidian/Utilities/ConsoleIO.cs
roxxel: image
From [readme](https://github.com/ORelio/Minecraft-Console-Client#readme)

Obsidian/ConsoleClient.cs
roxxel: This will force plugin devs to use try/catch blocks in commands which is not good ig

Obsidian/Program.cs
```diff
@@ -111,82 +111,76 @@ private static async Task Main()
 
         private static void InitConsoleInput()
         {
-            new Thread(new ThreadStart(async () =>
+            Task.Run(async () =>
             {
-                try
+                Server currentServer = Servers.First().Value;
+                Thread.Sleep(2000);
```
Seb-stian:
```cs
await Task.Delay(2000);
```

Obsidian/Commands/MainCommandModule.cs
```diff
@@ -272,14 +280,13 @@ public async Task TeleportAsync(CommandContext Context, [Remaining] VectorF loca
         [Command("op")]
         [CommandInfo("Give operator rights to a specific player.", "/op <player>")]
         [RequirePermission]
+        [IssuerScope(CommandIssuers.Any)]
         public async Task GiveOpAsync(CommandContext Context, IPlayer player)
         {
-            var server = (Server)Context.Server;
-            var onlinePlayers = server.OnlinePlayers;
-            if (!onlinePlayers.ContainsKey(player.Uuid) || !onlinePlayers.Any(x => x.Value.Username == player.Username))
+            if (player == null)
```
Naamloos, line 286: isn't player also null when issuer is the console? why have CommandIssuers.Any when consoles probably can't even execute it?

Obsidian/ConsoleClient.cs
Seb-stian: Well, they should check the value of `CommandContext.Issuer`, if they allow calling from the console. Perhaps `CommandIssuers.Client` should be the default, when `[IssuerScope]` is omitted? I'm not sure.

Naamloos: Exceptions are fine, devs should be aware of what can be done and what not.

Tides: Hmm this does look odd. I think there should be a new interface e.x ICommandSender for this instead of having ConsoleClient implement IPlayer ^ ^

Naamloos: in addition to that, returning `default` may return default values, giving a developer (or their code) a bad impression of what data is actually there. Just throw Exceptions like seb said

Naamloos: Adding even more to that, it might be a good idea adding abstractions to the Client class.

Obsidian/Utilities/ConsoleIO.cs
Naamloos: Good for now, but I'd prefer having this rewritten later on to avoid having weird third party licenses sprinkled in.

Obsidian/Utilities/Extensions.StringManipulation.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using Obsidian.API;
 using System;
+using System.Collections.Generic;
```
Naamloos: ...? Why did you add a single reference and nothing else?

Naamloos (changes requested):
LGTM, but please take another look at the comments made. Won't approve until further notice.

Obsidian/Commands/MainCommandModule.cs
roxxel: 🤨   `player` it's the player to whom the op, not an command issuer

Obsidian.API/_Enums/CommandsIssuer.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Obsidian.API
+{
+    [Flags]
+    public enum CommandIssuers
```
Craftplacer: File name does not match class name

Craftplacer (changes requested):
Almost there. Also, did you check up with giant list of overrides of the ConsoleClient?

Obsidian.API/_Enums/CommandsIssuer.cs
roxxel: > Almost there. Also, did you check up with giant list of overrides of the ConsoleClient?

Oh, I forgot to delete ConsoleClient, it's no longer needed

Craftplacer:
File names still not match, but at this point, I don't care

PR #104, opened 2021-05-15
Added TimeUpdate packet

Ma-tes:


Seb-stian (approved):
LGTM 🤠

Obsidian/Commands/MainCommandModule.cs
```diff
@@ -363,6 +367,23 @@ public async Task StopAsync(CommandContext Context)
...
+        #region time
+
+        [Command("time")]
+        [CommandInfo("Sets declared time", "/time <timeOfDay>")]
+        public async Task TimeAsync(CommandContext Context) => TimeAsync(Context, 1337);
+        [CommandOverload]
+        public async Task TimeAsync(CommandContext Context,int time)
+        {
+            var player = Context.Player as Player;
+            player.client.SendPacket(new TimeUpdate(0, time));
```
roxxel, line 381: I think this should be moved to IWorld to allow usage from plugins

Seb-stian: Good point! The command will be changed anyways, since commands are broken now (I will open an issue for it). It's more or less a test command right now.

Naamloos: I'd like to point out that the time packet here is not being broadcasted, it's only being sent to the executing client.

Naamloos: + if we merge it it might be nice to:
A. track time internally
B. send time to client when it joins.

Naamloos:
LGTM. Some comments on the command but IMO it can be merged as-is. Further specific behavior is out of scope for this specific PR.

Naamloos:
Also, thank you for your contribution!

Craftplacer (approved):
Not complete, but still a good start. LGTM.

PR #105, opened 2021-05-26
SetExperience packet

Ma-tes:


Seb-stian (approved):
Formatting is a little off, but otherwise looks great. More packets!

Obsidian/Net/Packets/Play/Clientbound/SetExperience.cs
```diff
@@ -0,0 +1,43 @@
...
+        public SetExperience(float experienceBar, int level, int totalExperience)
+        {
+            ExperienceBar = experienceBar;
+            Level = level;
+            TotalExperience = totalExperience;
+        }
+
+        public static SetExperience FromLevel(int level) => new (0, level, (int)GetExperience(level));
+        
+        private static float GetExperience(int level)
```
Craftplacer, line 27: This method shouldn't inside a packet class, instead inside a helper class or similar

Craftplacer (approved):
Same points as @Seb-stian, but still LGTM

PR #106, opened 2021-05-26
New UpdateHealth packet

Ma-tes:


Craftplacer (approved):
Formatting is off, please put attributes in their own lines and watch out that the indentation is correct.

Naamloos (changes requested):
Like craft said, this is not the way we do attributes in this repository. I'd appreciate it if you added a commit fixing this.

Seb-stian:
Added in 1529c321972dfd3f0b9de9586a5be77b549ad739 (breaking changes were pushed on master, so I fixed and added this packet). Thanks for your contribution!

PR #118, opened 2021-06-18
Feature/partial chunks

Jonpro03:
Implement Partial Chunks.
Introduce biome features and redo Tree generation as a feature.


Obsidian/Net/Packets/Play/Serverbound/PlayerBlockPlacement.cs
```diff
@@ -54,8 +56,9 @@ public async ValueTask HandleAsync(Server server, Player player)
                     Server.LastInventoryId = 1;
 
                 var maxId = Math.Max((byte)1, ++Server.LastInventoryId);
+                var bm = (BlockMeta)server.World.GetBlockMeta(position);
 
-                if (server.World.GetBlockMeta(position) is BlockMeta meta && meta.InventoryId != Guid.Empty)
+                if (bm is BlockMeta meta && meta.InventoryId != Guid.Empty)
```
Craftplacer: Unnecessary casting

Obsidian/Net/Packets/Play/Serverbound/PlayerBlockPlacement.cs
```diff
@@ -37,7 +37,9 @@ public async ValueTask HandleAsync(Server server, Player player)
 
             var position = this.Position;
 
-            var interactedBlock = server.World.GetBlock(position);
+            var b = server.World.GetBlock(position);
```
Craftplacer, line 40: What return type does server.World.GetBlock have? Because in the next statement you're explictly casting it to Block again, which might be unnecessary.

Craftplacer (changes requested):
Looks good to me for the most part, although at some points the code seems to have redundant statements.

Obsidian/Net/Packets/Play/Serverbound/PlayerBlockPlacement.cs
Jonpro03: GetBlock now returns a nullable block. This is to cast it from Block? to Block. [ref](https://github.com/ObsidianMC/Obsidian/pull/118/files/06ddba56639af82233220ad3a04be4312424b6f3#diff-07d01dc2121192313271a9b27d049be22f90141a5a9f21001f3839f2499ea90dR208)

Obsidian/Net/Packets/Play/Serverbound/PlayerBlockPlacement.cs
Jonpro03: GetBlockMeta returns a nullable type

Obsidian/Net/Packets/Play/Serverbound/PlayerBlockPlacement.cs
Craftplacer: Then I'd work with the property values `b.HasValue` and `b.Value`, but it's up to you if you wanna do that.

Obsidian/Net/Packets/Play/Serverbound/PlayerBlockPlacement.cs
Craftplacer: If you're checking the type in this if statement while also outputting a variable then it doesn't matter if it's nullable because it will be false if it's null.

```cs
                var bm = server.World.GetBlockMeta(position);
                // This won't be run if bm is null, and if it isn't null, meta is defined and not null.
                if (bm is BlockMeta meta && meta.InventoryId != Guid.Empty)
```

Craftplacer: FYI, the `is (type)` is doing type casting for you, `meta` is implictly set to the type you are checking against.

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/is

Jonpro03: Ahh, got it. I hadn't actually read through the code past fixing the build error :)

PR #120, opened 2021-07-03
Add editorconfig settings

Seb-stian:
In efforts to unify and enforce certain code style, we can add an `.editorconfig` file. These project settings take priority over local ones. We may want to discuss how it should be configured. (I kept pretty much everything as it was set by default).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,152 @@
+# Rules in this file were initially inferred by Visual Studio IntelliCode from the W:\Repository\csharp\net5\Obsidian codebase based on best match to current usage at 3.7.2021
+# You can modify the rules from these initially generated values to suit your own policies
+# You can learn more about editorconfig here: https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference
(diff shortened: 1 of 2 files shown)


Naamloos:
LGTM I guess. I haven't touched editor configs before, but unifying might be a good first step.

.editorconfig
```diff
@@ -0,0 +1,152 @@
+# Rules in this file were initially inferred by Visual Studio IntelliCode from the W:\Repository\csharp\net5\Obsidian codebase based on best match to current usage at 3.7.2021
+# You can modify the rules from these initially generated values to suit your own policies
+# You can learn more about editorconfig here: https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference
+[*.cs]
+
+
+#Core editorconfig formatting - indentation
+
+#use soft tabs (spaces) for indentation
+indent_style = space
```
Naamloos, line 10: Personally prefer tab, but that shouldn't matter too much :)

Naamloos (approved):
Looks good to me. Will merge.

PR #122, opened 2021-07-15
Changed BitConverter to BinaryPrimitives

MechSlayer:
Using BinaryPrimitives instead of BitConverter + reverse can speed up the the encoding/decoding of primitives from 300ns to 0.2ns


Craftplacer (approved):
LGTM

Seb-stian (approved):
LGTM. Although I would like MinecraftStream completely replaced in the future, this is good for now for sure :+1:

PR #127, opened 2021-07-28
Change MemoryWriter(ProtocolWriter) and implement MemoryReader(ProtocolReader)

MechSlayer:
Changed ProtocolWriter to use Span instead of array allowing us to use multiple types as a target buffer (array, stack spans, lists, etc...) this new version is more versatile but can be up to 80% slower than the old version depending on the method (writing 100 int64's takes 700ns in the new version and 400ns in the old one, writing 100 guid's takes 1100ns in the new version and 400ns in the old one) but can be optimized, it also grows automatically if required.


Seb-stian (commented):
I'm not very happy with the performance, but it brought up some good points.
Namely:
- EnsureCapacity on every call may have unnoticeable overhead
- Writer should keep track of the ArrayPool it has borrowed from
- We should look into whether BigEndianGuid is needed (probably is)

Buffer management seems to be improved overall, but since requirements are not final, it's not that critical (in future we may be able to avoid all copying).

Obsidian.IO/ProtocolReader.cs
```diff
@@ -0,0 +1,300 @@
...
+        private void EnsureSize(int required)
+        {
+            if (index + required > length)
+                throw new IndexOutOfRangeException("Cannot read past buffer's end");
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public byte ReadByte()
+        {
+            EnsureSize(1);
```
Seb-stian, line 105: Me: Hey mom, can we have `sizeof(byte)`?
Mom: No, we have `sizeof(byte)` at home
`sizeof(byte)` at home:

Obsidian.IO/ProtocolReader.cs
```diff
@@ -0,0 +1,299 @@
...
+        public string ReadString(int count)
+        {
+            if (count == 0) return string.Empty;
+
+            if (count <= 1024)
+            {
+                Span<byte> span = stackalloc byte[count];
+                ReadSpan(span);
+                return Encoding.UTF8.GetString(span);
+            }
```
Seb-stian, line 267:
```cs
System.Text.Encoding.UTF8.GetByteCount("☺") // → 3
```

This is not safe for all strings

Obsidian.IO/ProtocolReader.cs
```diff
@@ -0,0 +1,299 @@
...
+            var arrSpan = arr.AsSpan(0, count);
+            ReadSpan(arrSpan);
+            
+            var str = Encoding.UTF8.GetString(arrSpan);
+            pool.Return(arr);
+            return str;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public Guid ReadGuid()
```
Seb-stian, line 281: Reminder to self: check whether this works

Obsidian.IO/ProtocolReader.cs
```diff
@@ -0,0 +1,299 @@
...
+            var id = Unsafe.ReadUnaligned<Guid>(ref GetBufferRef());
+            index += 16;
+            return id;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void ReadInt64Array(Span<long> span)
+        {
+            for (var i = 0; i < span.Length; i++)
+                span[i] = ReadInt64();
```
Seb-stian, line 292: Isn't there some method for this? What comes to mind is `Span<byte> MemoryMarshal.AsBytes<long>(span)` and then `Span.CopyTo`, but there might be something even better.

Obsidian.IO/ProtocolWriter.cs
```diff
@@ -0,0 +1,418 @@
...
+            pool.Return(buffer);
+            buffer = newBuffer;
+        }
+        #endregion
+
+        #region Writing primitives
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void WriteByte(byte value)
+        {
+            EnsureCapacity(1);
```
Seb-stian: `sizeof(byte)`

Obsidian.IO/ProtocolWriter.cs
```diff
@@ -0,0 +1,418 @@
...
+
+        #region Writing commons
+        /// <summary>
+        /// Writes a <see cref="string"/> prefixed by VarInt.
+        /// </summary>
+        public void WriteVarString(string? value)
+        {
+            if (string.IsNullOrEmpty(value))
+            {
+                buffer[index++] = 0;
```
Seb-stian: `WriteByte(0)`

Obsidian.IO/ProtocolWriter.cs
```diff
@@ -0,0 +1,418 @@
...
+        }
+        
+        /// <summary>
+        /// Writes a <see cref="string"/> prefixed by <see cref="ushort"/>.
+        /// </summary>
+        public void WriteUShortString(string? value)
+        {
+            if (string.IsNullOrEmpty(value))
+            {
+                buffer[index++] = 0;
```
Seb-stian: `WriteByte(0)`

Obsidian.IO/ProtocolWriter.cs
```diff
@@ -0,0 +1,418 @@
...
+        }
+
+        public unsafe void WriteGuid(Guid value)
+        {
+            EnsureCapacity(16);
+            var ptr = (byte*)&value;
+            fixed (byte* bptr = buffer)
+            {
+                System.Buffer.MemoryCopy(ptr, bptr + index, buffer.Length, 16);
+            }
```
Seb-stian: We can probably use `Unsafe.WriteUnaligned` here

Obsidian.IO/ProtocolWriter.cs
```diff
@@ -0,0 +1,418 @@
...
+
+                if (unsigned != 0)
+                    temp |= 128;
+
+                ptr[size++] = temp;
+            } while (unsigned != 0);
+            
+            fixed (byte* bptr = buffer)
+                System.Buffer.MemoryCopy(ptr, bptr, buffer.Length, size);
+            index += size;
```
Seb-stian: The old version had some optimizations done to it. I wonder what the difference in performance is here.

Obsidian.IO/ProtocolReader.cs
Seb-stian: False alarm

Seb-stian:
I'd like to merge these changes.

PR #129, opened 2021-07-29
Fix encryption request packet, online mode is back bois

Naamloos:


Naamloos:
ah whatever I'll just merge, it's a small fix.

PR #145, opened 2021-09-04
Add ConcurrentMap

Seb-stian:
Adds new utility collection - `ConcurrentMap`. Internally it uses a `ReaderWriterLockSlim`, which should be more efficient in some cases, as it allows for multiple concurrent reads.

Possible issues:
- Adding more complexity → more possibilities for bugs (enumerations are difficult to get right)
- Name may be confusing, as it's hard to tell apart from `ConcurrentDictionary`

TODO
- [ ] Use for other often-read-rarely-changed dictionaries throughout codebase
- [ ] Test in action
- [ ] Add unit tests (?)

### DIFF
--- a/Obsidian/Server.cs
+++ b/Obsidian/Server.cs
@@ -15,6 +15,7 @@ using Obsidian.Net.Packets.Play.Clientbound;
 using Obsidian.Net.Packets.Play.Serverbound;
 using Obsidian.Plugins;
 using Obsidian.Utilities;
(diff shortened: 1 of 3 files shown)


Seb-stian:
`ConcurrentDictionary` is optimized for common reads and rare changes. The opposite of what I thought. In fact, the absolute chad who designed `ConcurrentDictionary` made its reads lock-less. This makes the draft itself useless, but I identified some other issues in our codebase along the way.

PR #156, opened 2021-10-23
Anonymized server listing options

Naamloos:
Add two config options controlling the server list: 
one of them disabled querying (drops connection as soon as state 1 (status) is detected
the other disabled the "user list sample" (player counts are still intact, but no players are listed)


Naamloos:
can someone review pls 🥺

Obsidian/Utilities/Config.cs
```diff
@@ -40,5 +40,9 @@ public class Config : IConfig
         public bool UDPBroadcast = false;
         
         public int PregenerateChunkRange { get; set; } = 15; // by default, pregenerate range from -15 to 15
+
+        public bool AnonymousOnlineList { get; set; } = false;
```
roxxel: maybe change this to enum?
like
public QueryingType Querying { get; set; }
```cs
enum QueryingType 
{
    Full,
    Anonymized,
    Disabled
}
```

Obsidian/Client.cs
```diff
@@ -149,11 +149,20 @@ public async Task StartConnectionAsync()
...
+                            this.Disconnect();
+                            break;
+                        }
                         switch (id)
                         {
                             case 0x00:
                                 var status = new ServerStatus(Server);
 
+                                if (config.AnonymousOnlineList) // If online list anonymous, clear the player sample.
+                                    status.Players.Sample.Clear();
```
Craftplacer: That the player list needs to be cleared seems odd, maybe a change inside ServerStatus might be worth to do.

Obsidian/Client.cs
```diff
@@ -149,11 +149,20 @@ public async Task StartConnectionAsync()
                 switch (this.State)
                 {
                     case ClientState.Status: // Server ping/list
+                        if(!this.config.EnableQuerying)
+                        {
+                            this.Logger.LogInformation("Closing connection, querying is disabled.");
```
Craftplacer: Maybe we shouldn't log this, only when in Verbose

Craftplacer (approved):

Obsidian/Client.cs
Naamloos: fair, will change

Naamloos:
Resolved all comments.

Craftplacer (approved):

PR #158, opened 2021-10-23
Remove tiddies

Naamloos:
The "tiddies" were a bit too much. The new text is a bit more tame.


Seb-stian (approved):
Censorship D:

Naamloos:
I'd love to keep it, but I feel like this may come over as a bit weird lol

to be fair, UwU is weird too, but less "graphic"

PR #159, opened 2021-10-24
Make MinecraftStream use ArrayPool for async methods

MechSlayer:
Though this can help with the GC pressure, because is using the Shared array pool it can end up using a lot of memory if there is a lot of traffic.

This could be solved by making each Client past the Status state have it's own ArrayPool.


Naamloos (approved):
Honestly, looks good to me. If this ends up being too heavy we can always decide on a per-client pool later on.

Seb-stian:
Looks good, except that those async methods should be made obsolete.

Craftplacer (approved):
LGTM

PR #162, opened 2021-10-27
Make object initialization consistent

### LINKED ISSUE #144: Make object initializers consistent.
When instantiating objects and setting members simultaneously, it can be seen that there are two ways of doing this (sometimes in the same file:
```
var x = new Something() { Foo = 3 };
var y = new Something { Foo = 3 };
```
Both are perfectly acceptable, but mixing and matching should be avoided.

Seb-stian:
Closes #144

I searched through the solution with regex and fixed all encountered inconsistencies.

### DIFF
--- a/Obsidian.API/_Types/ChatMessage.cs
+++ b/Obsidian.API/_Types/ChatMessage.cs
@@ -45,8 +45,8 @@ namespace Obsidian.API
         /// <summary>
         /// Creates a new <see cref="ChatMessage"/> object with plain text.
         /// </summary>
-        public static ChatMessage Simple(string text) => new ChatMessage() { Text = text.Replace('&', '§') };
-        public static ChatMessage Simple(string text, ChatColor color) => new ChatMessage() { Text = $"{color}{text}" };
+        public static ChatMessage Simple(string text) => new() { Text = text.Replace('&', '§') };
+        public static ChatMessage Simple(string text, ChatColor color) => new() { Text = $"{color}{text}" };
 
         public ChatMessage AddExtra(ChatMessage message)
         {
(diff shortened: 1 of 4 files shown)


Obsidian.API/_Types/ChatMessage.cs
```diff
@@ -45,8 +45,8 @@ public IEnumerable<ChatMessage> GetExtras()
         /// <summary>
         /// Creates a new <see cref="ChatMessage"/> object with plain text.
         /// </summary>
-        public static ChatMessage Simple(string text) => new ChatMessage() { Text = text.Replace('&', '§') };
-        public static ChatMessage Simple(string text, ChatColor color) => new ChatMessage() { Text = $"{color}{text}" };
+        public static ChatMessage Simple(string text) => new() { Text = text.Replace('&', '§') };
```
Naamloos, line 48: this uses new() { ...values }

Obsidian/Client.cs
```diff
@@ -90,7 +90,7 @@ public Client(TcpClient tcp, Config config, int playerId, Server originServer)
             Stream parentStream = this.tcp.GetStream();
             this.minecraftStream = new MinecraftStream(parentStream);
 
-            var blockOptions = new ExecutionDataflowBlockOptions() { CancellationToken = Cancellation.Token, EnsureOrdered = true };
+            var blockOptions = new ExecutionDataflowBlockOptions { CancellationToken = Cancellation.Token, EnsureOrdered = true };
```
Naamloos, line 93: this uses new ClassName { ...values }

Naamloos (changes requested):
ironically, in this PR alone there still seems to be inconsistencies. Look at the two comments I added. These both have different formats of object initialization.

Seb-stian:
> These both have different formats of object initialization.

Those are different cases though. With `ChatMessage`, type is implicit, thus can be left out. `new { ... }` is not valid (that would be an anonymous type), so `new() { ... }` is used.

Naamloos:
Fair, good point. Will accept and merge o/

Naamloos (approved):

PR #164, opened 2021-10-28
Make Obsidian support single server instance

ndelta0:
Lets face it, no one is going to be running multiple server instances on one executable if even machine.


ndelta0:
Well, since the #165 was merged there's no point for this PR to stick around.

PR #165, opened 2021-10-28
Remove multiserver functionality.

Naamloos:
This PR removes multi-world functionality, planning "official" bungeecord support in a future update(?)

This PR also changes the following:
- default world folder has been renamed `overworld`, in preparation for a future nether/end dimension.
- console-specific commands like `.switch` and `.execute` have been removed, as these were for multiserver support.
- folders and files that were once server-specific were moved to the main "obsidian" folder.
- Some globals have been moved to the server. Might be nice to move more globals to specific classes at some alter point, as statics are bad practice.
- Globalconfig is gone. Values have been moved to the regular config.


Naamloos:
Would be really cool if someone could review this one before the end of october... hacktober and all y'know? :D

Obsidian/Program.cs
```diff
@@ -58,66 +58,44 @@ private static async Task Main()
                 Console.CancelKeyPress += OnConsoleCancelKeyPressed;
             }
 
-            var globalConfigFile = new FileInfo(Program.globalConfigFile);
+            string serverDir = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
```
Craftplacer, line 61: This shouldn't use the binary path in any way, a relative path should be used instead because .NET will always resolve it to the current working directory.

Consider changing it to a `string.Empty` for now until we might have to change it for something else.

Craftplacer: This shouldn't use the binary path in any way, a relative path should be used instead because .NET will always resolve it to the current working directory.

Consider changing it to a `string.Empty` for now until we might have to change it for something else.

Craftplacer (changes requested):

Obsidian/Program.cs
Naamloos: Changes made.

Craftplacer (approved):

Issue #163, opened 2021-10-28
Resolve major warnings by GH actions compiler


Naamloos:
Builds generated by GitHub produce some warnings. these need to be fixed. Check the actions tab for output. Any xmldoc issues that nag about non-existant docs can be ignored for now.

HenryMigo:
I can pick this one up

HenryMigo:
It seems a lot of the warnings could be solved with constructors within the models. Could take a while to implement these constructors.

EDIT: It also looks like the build only shows the top 10 warnings. If this is correct then you've got around 3791 warnings to resolve. What I suggest is we eliminate these warnings in small batches to avoid thousands of changes. @Seb-stian do you agree with this approach?

Naamloos:
Do note that I was referring to the warnings given by github. Warnings by vs/the compiler are of course still welcome to be resolved.

HenryMigo:
> 
> 
> Do note that I was referring to the warnings given by github. Warnings by vs/the compiler are of course still welcome to be resolved.

It looks like it is the compilers warnings in VS but only the top 10 of these as if you drill into the build task in Actions it shows you all the warnings.

Naamloos:
> > 
> > 
> > Do note that I was referring to the warnings given by github. Warnings by vs/the compiler are of course still welcome to be resolved.
> 
> It looks like it is the compilers warnings in VS but only the top 10 of these as if you drill into the build task in Actions it shows you all the warnings.

Ahh, fair. I haven't looked into them that much yet 😅

HenryMigo:
I'll try and tackle as much as I can but don't want to touch too many files in one go.

HenryMigo:
Just a question, I am seeing async methods around the code that do not have awaitable code within them, do these need to be async and if so `await Task.Run(() => OUR_CODE);` will need to be used.

Seb-stian:
Yes, we do have fire-and-forget tasks here and there. The point is to have some code execute on a thread pool, while execution of the calling code continues.

Seb-stian:
Also sorry, didn't notice the edit before. That approach sounds reasonable.

HenryMigo:
I think I have managed to solve a few, most are XML warnings. 3692 relate to XML out of the 3708 while there are 16 other warnings. PR is #168

Naamloos:
I'll probably look into muting the "missing xmldoc" warnings project-wide sometime soon.

Issue #79, opened 2021-01-19
Further document Obsidian with XMLDocs


Naamloos:
This is especially useful for automatically generating API Documentation.

Assigning myself, but any help is welcome.

Naamloos:
Hey hacktoberfest people: this may or may not be a simple one for you! any documentation of undocumented methods is welcome.

PR #168, opened 2021-10-30
Compiler Warnings Clean up

HenryMigo:
Have done some clean-up within the code to remove compiler warnings.

I have implemented constructors where I can and ignore warnings as a last resort. In some cases we need to make sure where we don't allow NULL we have only constructors for that model, otherwise more NULL checking needs to happen within the code.

Outstanding Warnings:
- 3692 relate to XML out of the 3708
- 16 are other warnings

Relates to issue: #163

### DIFF
--- a/Obsidian.API/Crafting/SmithingRecipe.cs
+++ b/Obsidian.API/Crafting/SmithingRecipe.cs
@@ -14,13 +14,13 @@
 
         public Ingredient Addition { get; set; }
 
-        public SmithingRecipe(string name, CraftingType type, string? group, Ingredient result, Ingredient Base, Ingredient addition)
+        public SmithingRecipe(string name, CraftingType type, string? group, Ingredient result, Ingredient @base, Ingredient addition)
         {
             Name = name;
             Type = type;
             Group = group;
             Result = result;
-            this.Base = Base;
+            Base = @base;
             Addition = addition;
         }
     }
--- a/Obsidian.API/Crafting/SmithingRecipe.cs
+++ b/Obsidian.API/Crafting/SmithingRecipe.cs
@@ -13,5 +13,15 @@
         public Ingredient Base { get; set; }
 
         public Ingredient Addition { get; set; }
+
+        public SmithingRecipe(string name, CraftingType type, string? group, Ingredient result, Ingredient Base, Ingredient addition)
+        {
+            Name = name;
+            Type = type;
+            Group = group;
+            Result = result;
+            this.Base = Base;
+            Addition = addition;
+        }
     }
 }
--- a/Obsidian.API/Plugins/PluginAttribute.cs
+++ b/Obsidian.API/Plugins/PluginAttribute.cs
@@ -8,6 +8,14 @@ namespace Obsidian.API.Plugins
     [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
     public class PluginAttribute : Attribute
     {
+        /// <summary>
+        /// Plugin attribute constructor.
+        /// </summary>
+        /// <param name="name"></param>
+        /// <param name="version"></param>
+        /// <param name="authors"></param>
+        /// <param name="description"></param>
+        /// <param name="projectUrl"></param>
         public PluginAttribute(string name, string version, string authors, string description, string projectUrl)
         {
             Name = name;
@@ -20,26 +28,26 @@ namespace Obsidian.API.Plugins
         /// <summary>
         /// Name of the plugin.
         /// </summary>
-        public string Name { get; set; }
+        public string Name { get; }
 
         /// <summary>
         /// Version of the plugin. The string should contain the major, minor, <i>[build]</i>, and <i>[revision]</i> numbers, split by a period character ('.').
         /// </summary>
-        public string Version { get; set; }
+        public string Version { get; }
 
         /// <summary>
         /// Name(s) of the plugin's author(s).
         /// </summary>
-        public string Authors { get; set; }
+        public string Authors { get; }
 
         /// <summary>
         /// Description of the plugin.
         /// </summary>
-        public string Description { get; set; }
+        public string Description { get; }
 
         /// <summary>
         /// URL address of where the plugin is hosted.
         /// </summary>
-        public string ProjectUrl { get; set; }
+        public string ProjectUrl { get; }
     }
 }
--- a/Obsidian.API/Plugins/PluginAttribute.cs
+++ b/Obsidian.API/Plugins/PluginAttribute.cs
@@ -25,21 +25,21 @@ namespace Obsidian.API.Plugins
         /// <summary>
         /// Version of the plugin. The string should contain the major, minor, <i>[build]</i>, and <i>[revision]</i> numbers, split by a period character ('.').
         /// </summary>
-        public string Version { get; set; }
+        public string? Version { get; set; }
 
         /// <summary>
         /// Name(s) of the plugin's author(s).
         /// </summary>
-        public string Authors { get; set; }
+        public string? Authors { get; set; }
 
         /// <summary>
         /// Description of the plugin.
         /// </summary>
-        public string Description { get; set; }
+        public string? Description { get; set; }
 
         /// <summary>
         /// URL address of where the plugin is hosted.
         /// </summary>
-        public string ProjectUrl { get; set; }
+        public string? ProjectUrl { get; set; }
     }
 }
--- a/Obsidian.API/Plugins/PluginAttribute.cs
+++ b/Obsidian.API/Plugins/PluginAttribute.cs
@@ -8,6 +8,15 @@ namespace Obsidian.API.Plugins
     [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
     public class PluginAttribute : Attribute
     {
+        public PluginAttribute(string name, string version, string authors, string description, string projectUrl)
+        {
+            Name = name;
+            Version = version;
+            Authors = authors;
+            Description = description;
+            ProjectUrl = projectUrl;
+        }
+
         /// <summary>
         /// Name of the plugin.
         /// </summary>
--- a/Obsidian.API/Plugins/PluginAttribute.cs
+++ b/Obsidian.API/Plugins/PluginAttribute.cs
@@ -12,17 +12,9 @@ namespace Obsidian.API.Plugins
         /// Plugin attribute constructor.
         /// </summary>
         /// <param name="name"></param>
-        /// <param name="version"></param>
-        /// <param name="authors"></param>
-        /// <param name="description"></param>
-        /// <param name="projectUrl"></param>
-        public PluginAttribute(string name, string version, string authors, string description, string projectUrl)
+        public PluginAttribute(string name)
         {
             Name = name;
-            Version = version;
-            Authors = authors;
-            Description = description;
-            ProjectUrl = projectUrl;
         }
 
         /// <summary>
@@ -33,21 +25,21 @@ namespace Obsidian.API.Plugins
         /// <summary>
         /// Version of the plugin. The string should contain the major, minor, <i>[build]</i>, and <i>[revision]</i> numbers, split by a period character ('.').
         /// </summary>
-        public string Version { get; }
+        public string Version { get; set; }
 
         /// <summary>
         /// Name(s) of the plugin's author(s).
         /// </summary>
-        public string Authors { get; }
+        public string Authors { get; set; }
 
         /// <summary>
         /// Description of the plugin.
         /// </summary>
-        public string Description { get; }
+        public string Description { get; set; }
 
         /// <summary>
         /// URL address of where the plugin is hosted.
         /// </summary>
-        public string ProjectUrl { get; }
+        public string ProjectUrl { get; set; }
     }
 }
--- a/Obsidian.API/Plugins/PluginBase.cs
+++ b/Obsidian.API/Plugins/PluginBase.cs
@@ -20,6 +20,18 @@ namespace Obsidian.API.Plugins
 
         private Type typeCache;
 
+        public PluginBase(Type TypeCache)
+        {
+            typeCache = TypeCache;
+        }
+
+#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
+        public PluginBase()
+#pragma warning restore CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
+        {
+
+        }
+
         public void RegisterCommand(Action action)
         {
             registerSingleCommand(action);
@@ -31,7 +43,7 @@ namespace Obsidian.API.Plugins
         public object? Invoke(string methodName, params object[] args)
         {
             var method = GetMethod(methodName, args);
-            return method.Invoke(this, args);
+            return method?.Invoke(this, args);
         }
 
         /// <summary>
@@ -46,12 +58,12 @@ namespace Obsidian.API.Plugins
                 args = args.Take(parameterCount).ToArray();
             try
             {
-                var result = method.Invoke(this, args);
-                if (result is Task task)
-                    return task;
-                else if (result is ValueTask valueTask)
-                    return valueTask.AsTask();
-                return Task.FromResult(result);
+                    var result = method?.Invoke(this, args);
+                    if (result is Task task)
+                        return task;
+                    else if (result is ValueTask valueTask)
+                        return valueTask.AsTask();
+                    return Task.FromResult(result);
             }
             catch (Exception e)
             {
@@ -66,7 +78,7 @@ namespace Obsidian.API.Plugins
         public Task InvokeAsync(string methodName, params object[] args)
         {
             var method = GetMethod(methodName, args);
-            var result = method.Invoke(this, args);
+            var result = method?.Invoke(this, args);
             if (result is Task task)
                 return task;
             else if (result is ValueTask valueTask)
@@ -80,7 +92,7 @@ namespace Obsidian.API.Plugins
         public T? Invoke<T>(string methodName, params object[] args)
         {
             var method = GetMethod(methodName, args);
-            return (T?)method.Invoke(this, args);
+            return (T?)method?.Invoke(this, args);
         }
 
         /// <summary>
@@ -90,7 +102,7 @@ namespace Obsidian.API.Plugins
         public Task<T?> InvokeAsync<T>(string methodName, params object[] args)
         {
             var method = GetMethod(methodName, args);
-            var result = method.Invoke(this, args);
+            var result = method?.Invoke(this, args);
             if (result is Task<T?> task)
                 return task;
             else if (result is ValueTask<T?> valueTask)
@@ -137,7 +149,7 @@ namespace Obsidian.API.Plugins
                 args = args.Take(parameterCount).ToArray();
             try
             {
-                return method.Invoke(this, args);
+                return method?.Invoke(this, args);
             }
             catch (Exception e)
             {
@@ -175,7 +187,7 @@ namespace Obsidian.API.Plugins
             }
         }
 
-        private MethodInfo GetMethod(string methodName, object[] args)
+        private MethodInfo? GetMethod(string methodName, object[] args)
         {
             args ??= Array.Empty<object>();
             var types = new Type[args.Length];
--- a/Obsidian.API/Plugins/PluginBase.cs
+++ b/Obsidian.API/Plugins/PluginBase.cs
@@ -51,12 +51,12 @@ namespace Obsidian.API.Plugins
                 args = args.Take(parameterCount).ToArray();
             try
             {
-                    var result = method?.Invoke(this, args);
-                    if (result is Task task)
-                        return task;
-                    else if (result is ValueTask valueTask)
-                        return valueTask.AsTask();
-                    return Task.FromResult(result);
+                var result = method?.Invoke(this, args);
+                if (result is Task task)
+                    return task;
+                else if (result is ValueTask valueTask)
+                    return valueTask.AsTask();
+                return Task.FromResult(result);
             }
             catch (Exception e)
             {
@@ -230,6 +230,7 @@ namespace Obsidian.API.Plugins
                         break;
                     }
                 }
+
                 if (match)
                     return (method, parameters.Length);
             }
--- a/Obsidian.API/Plugins/PluginBase.cs
+++ b/Obsidian.API/Plugins/PluginBase.cs
@@ -20,16 +20,9 @@ namespace Obsidian.API.Plugins
 
         private Type typeCache;
 
-        public PluginBase(Type TypeCache)
-        {
-            typeCache = TypeCache;
-        }
-
-#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
         public PluginBase()
-#pragma warning restore CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
         {
-
+            typeCache ??= GetType();
         }
 
         public void RegisterCommand(Action action)
--- a/Obsidian.API/Plugins/PluginWrapper.cs
+++ b/Obsidian.API/Plugins/PluginWrapper.cs
@@ -7,9 +7,16 @@ namespace Obsidian.API.Plugins
     /// </summary>
     public abstract class PluginWrapper
     {
-        internal PluginBase plugin;
+        internal PluginBase Plugin;
 
+        public PluginWrapper(PluginBase plugin)
+        {
+            Plugin = plugin;
+        }
+
+#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
         public PluginWrapper()
+#pragma warning restore CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
         {
 
         }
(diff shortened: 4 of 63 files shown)


HenryMigo:
It ended up getting quite big so I drew the line here.

Obsidian.API/Crafting/SmithingRecipe.cs
```diff
@@ -13,5 +13,15 @@ public sealed class SmithingRecipe : IRecipe
         public Ingredient Base { get; set; }
 
         public Ingredient Addition { get; set; }
+
+        public SmithingRecipe(string name, CraftingType type, string? group, Ingredient result, Ingredient Base, Ingredient addition)
```
Seb-stian: I would prefer `@base` instead of `Base`.

Obsidian.API/Plugins/PluginAttribute.cs
```diff
@@ -8,6 +8,15 @@ namespace Obsidian.API.Plugins
     [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
     public class PluginAttribute : Attribute
     {
+        public PluginAttribute(string name, string version, string authors, string description, string projectUrl)
```
Craftplacer: Since the constructor is for an attribute, it's unnecessary, either:
- Make a constructor for initializing properties (also make them get-only, this is the issue with the introduction of this constructor)
- Remove the constructor again

Obsidian.API/Plugins/PluginBase.cs
```diff
@@ -20,6 +20,18 @@ public abstract class PluginBase
...
+            typeCache = TypeCache;
+        }
+
+#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
+        public PluginBase()
+#pragma warning restore CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
+        {
+
+        }
+
```
Craftplacer: The best approach to take here would've been writing a method that handles the null-checking where it sets `typeCache` to `GetType()` when `null` and returns the result, replacing most of the `typeCache ??= GetType()` calls. After that it would be safe to make `typeCache` nullable.

Obsidian.API/Plugins/Services/IDiagnoser.cs
```diff
@@ -31,7 +31,7 @@ public interface IDiagnoser : ISecuredService
         /// <param name="createWindow">Indicates whether to start the process in a new window.</param>
         /// <param name="useShell">Indicates whether to use the operating system shell to start the process.</param>
         /// <returns>A new <see cref="IProcess"/> that is associated with the process resource, or null if no process resource is started. </returns>
-        public IProcess StartProcess(string fileName, string arguments = null, bool createWindow = true, bool useShell = false);
+        public IProcess StartProcess(string fileName, string arguments = "", bool createWindow = true, bool useShell = false);
```
Craftplacer:
```suggestion
        public IProcess StartProcess(string fileName, string? arguments = null, bool createWindow = true, bool useShell = false);
```

Could've been made to `string?` since the implementing classes *should've* supported null cases before.

Obsidian.API/Plugins/Services/ILogger.cs
```diff
@@ -53,7 +53,14 @@ public void LogTrace(string message)
...
         {
             Log($"{nameof(T)}: {exception.Message}");
-            Log(exception.StackTrace);
+            if (exception.StackTrace != null)
+            {
+                Log(exception.StackTrace);
+            }
+            else
+            {
+                Log("No stack trace...");
```
Craftplacer: Nit: Log message *can* be omitted

Obsidian.API/_Attributes/RequirePermissionAttribute.cs
```diff
@@ -19,6 +19,11 @@ public RequirePermissionAttribute(PermissionCheckType checkType = PermissionChec
 
         public override Task<bool> RunChecksAsync(CommandContext context)
         {
+            if (context.Player == null)
+            {
+                throw new Exception("Player not found");
```
Craftplacer: Not sure if this makes sense here since the possibility that the `Player` is `null` can always happen and might be expected when the console executes a command which uses this check. Also meaning that this code is put in the wrong place, instead make it return false like the code below and put it under the [console issuer check](https://github.com/ObsidianMC/Obsidian/pull/168/files#diff-8cda4a03e94a2cc8a5b3572e4351d6821f7b78359d098a1075bff1be6caf626eR27-R28).

Obsidian.API/_Types/Inventory/ItemMetaBuilder.cs
```diff
@@ -28,11 +28,11 @@ public class ItemMetaBuilder
 
         private readonly List<ChatMessage> lore = new List<ChatMessage>();
 
-        public ItemMetaBuilder()
+        public ItemMetaBuilder(string name = "")
         {
             this.Enchantments = new ReadOnlyDictionary<EnchantmentType, Enchantment>(this.enchantments);
             this.StoredEnchantments = new ReadOnlyDictionary<EnchantmentType, Enchantment>(this.storedEnchantments);
-
+            this.Name = name;
```
Craftplacer: Not sure about passing down an empty string as default argument to a property.

Obsidian/Utilities/RentedArray.cs
```diff
@@ -27,13 +27,19 @@ namespace Obsidian.Utilities
         /// </summary>
         public Memory<T> Memory => array.AsMemory(0, Length);
 
+#nullable enable
+        /// <summary>
+        /// #nullable
+        /// </summary>
+        /// <param name="length"></param>
+        /// <param name="pool"></param>
```
Craftplacer: Unnecessary XMLdoc.

Obsidian/WorldData/BlockUpdates.cs
```diff
@@ -9,171 +9,177 @@ internal static class BlockUpdates
...
+                var location = blockUpdate.position;
+                var material = blockUpdate.block.Value.Material;
+                if (world.GetBlock(location + Vector.Down) is Block below && (Block.Replaceable.Contains(below.Material) || below.IsFluid))
+                {
+                    world.SetBlock(location, Block.Air);
+                    world.SpawnFallingBlock(location, material);
+                    return true;
+                }
+                return false;
+            });
```
Craftplacer: I'm not sure if this should run in its own thread running possibly minor sync code just to solve an async warning.

Obsidian/WorldData/RegionFile.cs
```diff
@@ -233,7 +233,7 @@ protected virtual void Dispose(bool disposing)
             {
                 if (disposing)
                 {
-                    FlushToDiskAsync();
+                    Task.Run(async () => await FlushToDiskAsync());
```
Craftplacer: Should've used [IAsyncDisposable](https://docs.microsoft.com/en-us/dotnet/api/system.iasyncdisposable) in the first place.

Obsidian/WorldData/World.cs
```diff
@@ -529,9 +529,12 @@ public async Task ManageChunksAsync()
...
         {
-            Server.BroadcastMessage("Saving to disk...");
-            Parallel.ForEach(Regions.Values, async r => await r.FlushAsync());
-            Server.BroadcastMessage("Save complete.");
+            await Task.Run(() =>
+            {
+                Server.BroadcastMessage("Saving to disk...");
+                Parallel.ForEach(Regions.Values, async r => await r.FlushAsync());
+                Server.BroadcastMessage("Save complete.");
+            });
```
Craftplacer: Should possibly use [Task.WhenAll](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.whenall) instead, but I am unsure whether IO blocking could interfere when flushing

Craftplacer (changes requested):
Hey @HenryMigo, congrats on your first pull request.

I've reviewed it and found several issues, I'd like you to take a look at them and resolve them with us.

Due to the size of the pull request I'm going to request reviews from some of our team members.

Craftplacer:
> * 3692 relate to XML out of the 3708

Tracked in #79

Obsidian.API/Plugins/PluginAttribute.cs
HenryMigo: Hi Craft,

I have kept the constructor for initializing the properties and made the properties get only. Hope that's what you were asking.

Obsidian.API/Plugins/PluginBase.cs
HenryMigo: Have altered the empty constructor to `typeCache ??= GetType()` this now means the other one is not needed.

Obsidian.API/Plugins/Services/IDiagnoser.cs
HenryMigo: Good spot, I did think of this when I made the change but don't know why I thought against it.

HenryMigo: Having looked at this further the `ProcessStartInfo` which needs the values does NOT support nullable and this can't be changed to support them as it's not our code but Microsofts, I suggest we either follow suit or do a check for nullable and make it an empty string. @Craftplacer

Obsidian.API/Plugins/Services/ILogger.cs
HenryMigo: Have removed this else statement

Obsidian.API/Plugins/PluginAttribute.cs
Craftplacer: Alright, then this change is qualified. 👍

Obsidian.API/_Attributes/RequirePermissionAttribute.cs
HenryMigo: I have amended this, good spot, and thank you for the context!

Obsidian/Utilities/RentedArray.cs
HenryMigo: Correct, this looks like I left some code here which should've been cleaned up.

Obsidian/WorldData/RegionFile.cs
HenryMigo: I have implemented IAsyncDisposable and made the changes to the dispose method

Obsidian/WorldData/World.cs
HenryMigo: Implemented the WhenAll

HenryMigo:
I have made the changes as per suggestions, let me know what you think

Obsidian/WorldData/BlockUpdates.cs
HenryMigo: I have made this sync again, this will have the async warning but as you said I don't think it's too important to resolve this with this.

Obsidian.API/_Types/Inventory/ItemMetaBuilder.cs
HenryMigo: I have removed the constructor and left it as-is for now, the warning will be back for it though.

Obsidian.API/Plugins/Services/IDiagnoser.cs
HenryMigo: I have implemented in the concrete class to default to `string.Empty` if it finds NULL, had a look within the ProcessStartInfo class and it looks like that's what Microsoft is doing themselves.

Craftplacer (approved):

Obsidian.API/Plugins/PluginBase.cs
```diff
@@ -46,12 +51,12 @@ public Task FriendlyInvokeAsync(string methodName, params object[] args)
                 args = args.Take(parameterCount).ToArray();
             try
             {
-                var result = method.Invoke(this, args);
-                if (result is Task task)
-                    return task;
-                else if (result is ValueTask valueTask)
-                    return valueTask.AsTask();
-                return Task.FromResult(result);
+                    var result = method?.Invoke(this, args);
```
Naamloos: not sure why spacing went all crazy here, but I'll ignore this for now.

Obsidian.Tests/Noise.cs
```diff
@@ -13,37 +14,40 @@ public class Noise
...
 
-            var map = new NoiseMap();
+                var map = new NoiseMap();
 
-            PlaneNoiseMapBuilder builder = new PlaneNoiseMapBuilder()
-            {
-                DestNoiseMap = map,
-                SourceModule = noiseGen.Result
-            };
+                PlaneNoiseMapBuilder builder = new PlaneNoiseMapBuilder()
```
Naamloos: more funky spacing goofs

Obsidian/Utilities/RentedArray.cs
```diff
@@ -27,13 +27,14 @@ namespace Obsidian.Utilities
...
 
+#nullable enable
         public RentedArray(int length, ArrayPool<T>? pool = null)
         {
             this.pool = pool ?? ArrayPool<T>.Shared;
             array = this.pool.Rent(length);
             Length = length;
         }
-        
+#nullable disable
```
Naamloos: I am fairly sure nullables are (going to be?) enabled project-wide, so I am not 100% sure about whether this is necessary at all.

Obsidian/WorldData/RegionFile.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace Obsidian.WorldData
 {
-    public class RegionFile : IDisposable
+    public class RegionFile : IAsyncDisposable
```
Naamloos, line 11: I was not aware of this being a thing but I am very glad that it is. Nice.

SamplePlugin/SamplePlugin.cs
```diff
@@ -7,9 +7,9 @@
 
 namespace SamplePlugin
 {
-    [Plugin(Name = "Sample Plugin", Version = "1.0",
-            Authors = "Obsidian Team", Description = "My sample plugin.",
-            ProjectUrl = "https://github.com/Naamloos/Obsidian")]
```
Naamloos, line 12: These values were chosen to be explicit for the sample plugin. If possible, please keep these explicit.

SamplePlugin/SamplePlugin.cs
```diff
@@ -19,7 +19,7 @@ public class SamplePlugin : PluginBase
         // Dependencies will be injected automatically, if dependency class and field/property names match
         // Plugins won't load until all their required dependencies are added
         // Optional dependencies may be injected at any time, if at all
-        [Dependency(MinVersion = "2.0", Optional = true), Alias("Sample Remote Plugin")]
```
Naamloos, line 22: Same here for previous comment. These were explicit for clarity.

Naamloos (approved):
I have some small changes I'd like to see. I added some comments. I'll approve the PR because hacktober is ending soon but consider this a "request changes".

Obsidian/WorldData/World.cs
```diff
@@ -529,11 +529,13 @@ public async Task ManageChunksAsync()
...
-                Server.BroadcastMessage("Saving to disk...");
-                Parallel.ForEach(Regions.Values, async r => await r.FlushAsync());
-                Server.BroadcastMessage("Save complete.");
+            await Task.WhenAll(new List<Task> {
+                Task.Run(() =>
+                {
+                    Server.BroadcastMessage("Saving to disk...");
+                    Parallel.ForEach(Regions.Values, async r => await r.FlushAsync());
+                    Server.BroadcastMessage("Save complete.");
+                })
```
Tides: You could just do `await Task.WhenAll(Regions.Values.Select(r => r.FlushAsync()))` and move the broadcast message back outside ^ ^

HenryMigo:
I will address the comments on this PR this evening if that's alright with everyone?

Naamloos:
> 
> 
> I will address the comments on this PR this evening if that's alright with everyone?

sure, no pressure

SamplePlugin/SamplePlugin.cs
HenryMigo: In the next commit, they will be explicit.

SamplePlugin/SamplePlugin.cs
HenryMigo: Same with this

Obsidian.API/Plugins/PluginBase.cs
HenryMigo: Resolved spacing in next commit

Obsidian.Tests/Noise.cs
HenryMigo: Resolved in next commit

Obsidian/Utilities/RentedArray.cs
HenryMigo: It unfortunately isn't enabled project-wide. I have enabled it project-wide in next commit

Obsidian.API/Plugins/PluginWrapper.cs
```diff
@@ -7,51 +7,58 @@ namespace Obsidian.API.Plugins
     /// </summary>
     public abstract class PluginWrapper
     {
-        internal PluginBase plugin;
+        internal PluginBase Plugin;
```
Seb-stian: Fields should be named with camelCase.

Obsidian.API/Plugins/Services/IFileWriter.cs
```diff
@@ -92,14 +105,30 @@ public void CopyFile(string sourceFileName, string destinationFileName)
             if (!IsUsable)
                 throw new SecurityException(SecurityExceptionMessage);
 
-            var workingDirectory = GetWorkingDirectory();
-            if (!Path.GetDirectoryName(Path.GetFullPath(sourceFileName)).StartsWith(workingDirectory, true, null) && Path.IsPathFullyQualified(sourceFileName)) throw new UnauthorizedAccessException(sourceFileName);
+            string workingDirectory = GetWorkingDirectory();
+            var fullPath1 = Path.GetDirectoryName(Path.GetFullPath(sourceFileName));
+            var fullPath2 = Path.GetDirectoryName(Path.GetFullPath(destinationFileName));
```
Seb-stian: I would prefer names like `sourceFilePath` and `destinationFilePath`.

Obsidian.API/Plugins/Services/IFileWriter.cs
```diff
@@ -114,13 +143,29 @@ public void MoveFile(string sourceFileName, string destinationFileName)
                 throw new SecurityException(SecurityExceptionMessage);
 
             var workingDirectory = GetWorkingDirectory();
-            if (!Path.GetDirectoryName(Path.GetFullPath(sourceFileName)).StartsWith(workingDirectory, true, null) && Path.IsPathFullyQualified(sourceFileName)) throw new UnauthorizedAccessException(sourceFileName);
+            var fullPath1 = Path.GetDirectoryName(Path.GetFullPath(sourceFileName));
+            var fullPath2 = Path.GetDirectoryName(Path.GetFullPath(destinationFileName));
```
Seb-stian: I would prefer names like sourceFilePath and destinationFilePath.

Obsidian.API/_Interfaces/ILiving.cs
```diff
@@ -4,7 +4,7 @@ public interface ILiving : IEntity
     {
         public LivingBitMask LivingBitMask { get; set; }
 
-        public float Health { get; set; }
+        public new float Health { get; set; }
```
Seb-stian: Hiding base members with `new` is a big nono. This will lead to a lot of unwanted behavior.

Same goes for `IPlayer.Uuid`.

Obsidian.API/_Types/ChatMessage.cs
```diff
@@ -6,6 +6,28 @@ namespace Obsidian.API
 {
     public class ChatMessage
     {
+        public ChatMessage(string text, HexColor color, bool bold, bool italic, bool underlined, bool strikethrough, bool obfuscated, string insertion, ClickComponent clickEvent, HoverComponent hoverEvent, List<ChatMessage> extra)
```
Seb-stian: I'm not sure if this the way to go for `ChatMessage`.

Obsidian/WorldData/World.cs
```diff
@@ -529,9 +529,7 @@ public async Task ManageChunksAsync()
 
         public async Task FlushRegionsAsync()
         {
-            Server.BroadcastMessage("Saving to disk...");
-            Parallel.ForEach(Regions.Values, async r => await r.FlushAsync());
-            Server.BroadcastMessage("Save complete.");
+            await Task.WhenAll(Regions.Values.Select(r => r.FlushAsync()));
```
Seb-stian: `await Task.WhenAll(Regions.Select(pair => pair.Value.FlushAsync()));`
It may not be as pretty, but `ConcurrentDictionary.Values` allocates a new list whenever called.

Obsidian/WorldData/BlockUpdates.cs
```diff
@@ -10,170 +10,183 @@ internal static class BlockUpdates
...
-            var world = blockUpdate.world;
-            var location = blockUpdate.position;
-            int state = block.State;
-            Vector belowPos = location + Vector.Down;
-            
-            // Handle the initial search for closet path downwards.
-            // Just going to do a crappy pathfind for now. We can do
-            // proper pathfinding some other time.
-            if (state == 0)
+            return await Task.Run(() =>
```
Seb-stian: `await` may be left out (and then method's `async`).

Obsidian.API/Plugins/PluginAttribute.cs
Seb-stian: I'm not a fan of making all these properties required via a constructor. What do the others think? @Tides

Obsidian.API/Plugins/Services/IFileWriter.cs
HenryMigo: Yep was meant to rename these after

Obsidian.API/Plugins/Services/IFileWriter.cs
HenryMigo: Renaming

Obsidian.API/Plugins/PluginWrapper.cs
HenryMigo: I was corrected by the IDE to use an upper case for non-private properties, will change it back.

Obsidian.API/Crafting/SmithingRecipe.cs
HenryMigo: have changed

Obsidian.API/_Interfaces/ILiving.cs
HenryMigo: have corrected

Obsidian/WorldData/World.cs
HenryMigo: updated to this

Obsidian/WorldData/BlockUpdates.cs
HenryMigo: Unsure what you mean here? Do you want me to make the method sync? As it'll break the code which is calling it with an await.

Obsidian.API/_Types/ChatMessage.cs
HenryMigo: I will remove and set this back to normal, but the compiler warnings will be back.

Obsidian/WorldData/BlockUpdates.cs
Seb-stian: I don't see why it should break; it would synchronously return a task, which would then be asynchronously awaited from some asynchronous method. That seems completely valid to me, but correct me if I'm wrong.

HenryMigo: Oh right, that's what it is doing before, so either we have the async declaration on the method or we do `Task.FromResult(false)`, is returning Task.FromResult okay? @Seb-stian

Seb-stian: Yes, that would be fine.

HenryMigo: Changed in commit

Seb-stian (approved):
Job well done.

HenryMigo:
Just wanted to check that someone is able to complete this PR?

HenryMigo:
@Craftplacer how do I get this PR completed?

Craftplacer (changes requested):
I'm almost satisfied with the changes, the only issue I still see is that the constructor still requires all values to be set ([related to Naam's review](https://github.com/ObsidianMC/Obsidian/pull/168#discussion_r739875737)).

I'd suggest making that the only constructor only requires and sets Name, as to the other values can be set using `[Plugin(name, Version = "1.0.0")]` etc.

HenryMigo:
> I'm almost satisfied with the changes, the only issue I still see is that the constructor still requires all values to be set ([related to Naam's review](https://github.com/ObsidianMC/Obsidian/pull/168#discussion_r739875737)).
> 
> I'd suggest making that the only constructor only requires and sets Name, as to the other values can be set using `[Plugin(name, Version = "1.0.0")]` etc.

I can make that change but this will bring back the compiler warning asking us to change the props to nullable if they are not in the constructor.

Craftplacer:
> > I'm almost satisfied with the changes, the only issue I still see is that the constructor still requires all values to be set ([related to Naam's review](https://github.com/ObsidianMC/Obsidian/pull/168#discussion_r739875737)).
> > 
> > I'd suggest making that the only constructor only requires and sets Name, as to the other values can be set using `[Plugin(name, Version = "1.0.0")]` etc.
> 
> I can make that change but this will bring back the compiler warning asking us to change the props to nullable if they are not in the constructor.

You should make the properties that aren't required nullable, let me know about any issues you encounter doing that.

HenryMigo:
> > > I'm almost satisfied with the changes, the only issue I still see is that the constructor still requires all values to be set ([related to Naam's review](https://github.com/ObsidianMC/Obsidian/pull/168#discussion_r739875737)).
> > > I'd suggest making that the only constructor only requires and sets Name, as to the other values can be set using `[Plugin(name, Version = "1.0.0")]` etc.
> > 
> > 
> > I can make that change but this will bring back the compiler warning asking us to change the props to nullable if they are not in the constructor.
> 
> You should make the properties that aren't required nullable, let me know about any issues you encounter doing that.

Changes made :)

Craftplacer (approved):
Aight

PR #169, opened 2021-10-31
Implement Basic Version of PaperMC Patch 0284

Aaron2550:
Patch 0284 Prevents Players from moving into unloaded Chunks
I did this in Entity.cs so all Entities cannot move into unloaded Chunks

I havent tested it yet because its 23:30 and i wanna sleep, I can test this tomorrow but in my head it has a chance of **_atleast_** 20% of working

Feel free to reject this PR if yall wanna go with a patch-like System like Czompi proposed or ... if the code is just shit i guess

### DIFF
--- a/Obsidian/Entities/Entity.cs
+++ b/Obsidian/Entities/Entity.cs
@@ -169,13 +169,27 @@ namespace Obsidian.Entities
 
         public void UpdatePosition(VectorF pos, bool onGround = true)
         {
-            this.Position = pos;
+            using (Chunk chunk = this.World.GetChunk(pos.X / 16f, pos.Y / 16f, false))
+            {
+                if (chunk != null && chunk.isGenerated)
+                {
+                    this.Position = pos;
+                }
+            }
+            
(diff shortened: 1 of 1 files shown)


Naamloos:
I mean- I can't accept this yet anyway as it fails to build..

Aaron2550:
Hell yeah I forgot to do the most basic of testing.... I guess those 20% did fail me. I'll check this later today

Aaron2550:
I just assumed its fine because VS didnt show me any errors, it didnt even underline it. I havent checked yet but my guess is that Chunk is not disposable

Aaron2550:
Once enderpearls are implemented, this may fix https://2b2t.miraheze.org/wiki/Ender_pearl_loading
Because once the Horizontal Velocity is 0, it'll just drop straight down... right?

Naamloos (approved):
Looks fine to me, makes sense ignoring the fact that this fixes a bug for a feature that isn't even introduced yet :p

Issue #170, opened 2021-11-01
Plugin creation too complicated


ndelta0:
In my opinion the current plugin system is way to complex for what it needs to be. There's the actual plugin implementation, the plugin wrapper, dependency injection and plugin permissions.

Regarding plugin permissions and controlling what they can and what they can't do is not possible currently. It was possible in older version of .NET Framework, but .NET Core and newer versions don't care about those rescrictions thus rendering the permissions obsolete.

I have a ready working new plugin implementation in a test solution, I was just waiting for #165 to be merged since it would greatly simplify the redesign.

Seb-stian:
> Regarding plugin permissions and controlling what they can and what they can't do is not possible currently. It was possible in older version of .NET Framework, but .NET Core and newer versions don't care about those rescrictions thus rendering the permissions obsolete.

We already discussed this.

I can't imagine how we could make, for example dependency injection simpler. Or do you not want these features at all? I'm all for simpler plugin system, as long as it doesn't make them too simple to the point of confusion (so much background magic and implicity that we no longer know what's really going on). And so, could you elaborate further on what exact changes you would like to see, so that we can discuss them?

ndelta0:
> I can't imagine how we could make, for example dependency injection simpler. Or do you not want these features at all?

No, I still want those features since they are essential for the way plugins would interact with each other.

I planned on ditching the Dependency Injection system an using a Service Provider. With this approach the plugin managed does not have to manage what plugin needs what things injected, but instead the plugin retrieves the required parts/"services" when it wants them, also allowing for the services to be optional if not found. On the other hand, each plugin can "export" their services at will, removing the part where whole plugins would get injected into other plugins which was unneccessarry.

Having services also allows for offloading things like a global database/ORM service which would greatly reduce the complexity of plugins which would require access to the DB. This would be also safe in way that the service could allow the plugin to interact with tables only for that plugin, controlled access to the other plugins data would be possible via the services those plugins would provide.

Seb-stian:
Plugins exporting services instead of being services themselves is not a bad idea. I wonder how it will turn out. Using Service Provider was already in scope of #111 , but it should be fine if implemented elsewhere.

ndelta0:
I see, though I looked at using those providers, but they are tailored specifically for ASP.NET and it would be to much hassle to make them work than creating a simple custom service provider.

Tides:
[...]

PR #175, opened 2021-11-04
Plugin framework overhaul

ndelta0:
Changed many aspects of the plugin system. Got rid of those bases, wrappers, etc. No command support **yet**, but it will come with a command system overhaul.

Ultimately I decided to create own service provider instead of the default one because of custom implementation being way easier and more tailored for the specific requirements.


ndelta0:
(Mentioning #170)

Seb-stian:
Some changes weren't agreed on after discussions with @Tides. Closing for now, may be used for future reference (regarding services & events).

PR #181, opened 2021-11-13
Feature/more biomes

Jonpro03:
* feat: Add Flora Decorator feature.
* fix: Frozen rivers only half frozen when bordering a warmer biome.
* feat: Clean up mountains biomes to make them look a tiny bit more realistic...


Naamloos (approved):
Honestly, what can I say besides LGTM?

PR #184, opened 2021-11-14
Use ArgumentNullException.ThrowIfNull

### LINKED ISSUE #173: Refactor throwing ArgumentNullException
With .NET 6, this code

```cs
if (foo is null)
    throw new ArgumentNullException(nameof(foo));
```

can be replaced by

```cs
ArgumentNullException.ThrowIfNull(foo, nameof(foo));
```

which is to be prefered.

Seb-stian:
Closes #173


Naamloos (approved):
fkn hot

Tides (approved):
LGTM

PR #188, opened 2021-11-22
Cleanup MinecraftAPI.cs

OoLunar:
Cleans up MinecraftAPI.cs by using the Json extension methods. Additionally marks some return types as nullable

### DIFF
--- a/Obsidian/Utilities/Mojang/MinecraftAPI.cs
+++ b/Obsidian/Utilities/Mojang/MinecraftAPI.cs
@@ -1,49 +1,44 @@
-﻿using System.Net.Http;
+using System.Net.Http;
+using System.Net.Http.Json;
 using System.Text;
-using System.Text.Json;
 using System.Web;
 
 namespace Obsidian.Utilities.Mojang;
 
 public class MinecraftAPI
 {
-    private static readonly HttpClient Http = Globals.HttpClient;
+    private static readonly HttpClient httpClient = new();
 
-    public static async Task<List<MojangUser>> GetUsersAsync(string[] usernames)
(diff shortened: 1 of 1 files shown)


Naamloos (approved):
LGTM, don't see anything weird.

OoLunar:
The httpClient methods used could optionally be modified to allow a CancellationToken, however that'd require modifying the method parameters and would thus be a breaking change

Tides (changes requested):
The json extension methods should take in our `Globals.JsonOptions` for the serializer options param (To avoid any de/serialization problems)

Seb-stian (approved):
I suppose we should make those string literals `static readonly`, but that may be beyond the scope of this PR. I like the changes.

Obsidian/Utilities/Mojang/MinecraftAPI.cs
```diff
@@ -1,49 +1,43 @@
...
 using System.Text;
-using System.Text.Json;
 using System.Web;
 
 namespace Obsidian.Utilities.Mojang;
 
 public class MinecraftAPI
 {
-    private static readonly HttpClient Http = Globals.HttpClient;
+    private static readonly HttpClient httpClient = new();
```
Craftplacer: Why is a new instance made inside `MinecraftAPI`?

OoLunar: Missed that. Meant to use the Global HttpClient when adding the Global Json Options

OoLunar: Fixed as of commit 63905f2

Craftplacer (approved):

PR #210, opened 2021-12-29
Remove memory copies when loading chunks

Seb-stian:
Removes allocations coming from `byte[] Memory<byte>.ToArray()`

### DIFF
--- a/Obsidian/Utilities/ReadOnlyStream.cs
+++ b/Obsidian/Utilities/ReadOnlyStream.cs
@@ -1,0 +1,74 @@
+﻿using System.IO;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace Obsidian.Utilities;
+
+internal sealed class ReadOnlyStream : Stream
+{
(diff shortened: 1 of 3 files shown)


Obsidian/Utilities/ReadOnlyStream.cs
```diff
@@ -0,0 +1,74 @@
...
+        return MemoryMarshal.CreateSpan(ref first, size);
+    }
+
+    public override long Seek(long offset, SeekOrigin origin)
+    {
+        Position = origin switch
+        {
+            SeekOrigin.Begin => offset,
+            SeekOrigin.Current => index + offset,
+            SeekOrigin.End => length + offset,
```
roxxel, line 65: Shouldn't it be `length - offset`?

Seb-stian: No, it is expected for `offset` to be a negative number. This is common behaviour for .NET streams.

Jonpro03 (approved):
LGTM

Naamloos (approved):

PR #215, opened 2022-01-03
Surpress warnings that we're not going to fix anyway

Naamloos:
Let me know what you think.


Seb-stian:
I would like to fix them instead :d

Naamloos:
ba okay brb

PR #219, opened 2022-01-05
Various fixes

roxxel:
# Summary 
Fixed op/deop command (in console execution throws null ref ex)
Added `IsPlayer` property to CommandContext
Replaced Context.Player to Context.Sender in every command that not restricted for player only execution 
Simplified check in gamemode command

### DIFF
--- a/Obsidian/Commands/MainCommandModule.cs
+++ b/Obsidian/Commands/MainCommandModule.cs
@@ -199,7 +199,7 @@ public class MainCommandModule
     [Command("uptime", "up")]
     [CommandInfo("Gets current uptime", "/uptime")]
     public Task UptimeAsync(CommandContext Context)
-        => Context.Player.SendMessageAsync($"Uptime: {DateTimeOffset.Now.Subtract(Context.Server.StartTime)}");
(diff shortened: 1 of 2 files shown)


Obsidian/Commands/MainCommandModule.cs
```diff
@@ -224,7 +224,7 @@ public async Task GamemodeAsync(CommandContext Context, [Remaining] string args_
         var args = args_.Contains(" ") ? args_.Split(" ").ToList() : new List<string> { args_ };
         if (args.Count == 1)
         {
-            if (args[0].ToLower() == "creative" || args[0].ToLower() == "survival" || args[0].ToLower() == "spectator" || args[0].ToLower() == "adventure")
+            if (new[] { "creative", "survival", "spectator", "adventure" }.Contains(args[0].ToLowerInvariant()))
```
Seb-stian, line 227: Allocating an array for each call is not ideal, but it's still better than what was there previously.

Seb-stian (approved):
LGTM

Naamloos (approved):

PR #220, opened 2022-01-07
Docker support

Naamloos:


Naamloos:
one review is fine, just requesting anyone for this small PR :))

.dockerignore
```diff
@@ -0,0 +1,4 @@
+.github
+.gitignore
+appveyor.yml
+azure-pipelines.yml
```
Seb-stian: Those files no longer exist

Naamloos: oh right thanks. resolving that before merging.

PR #227, opened 2022-01-09
Basic nick/ip whitelist

### LINKED ISSUE #224: Network-based whitelisting
As suggested in #153, IP whitelists might be nice to have. This means immediately blocking when an incoming IP is not whitelisted, not even advertising the fact the port belongs to a minecraft server (unless you're whitelisted, of course.)

priority is low, but it's really nice to have some low level whitelisting

roxxel:
Resolves #224 
# Summary
- Server drops connection if ip is not whitelisted
- Server kicks player on connection if nickname is not whitelisted
- ~~Commands for managing whitelist~~ postponed until command framework rework (https://discord.com/channels/772894170451804220/772894226756403251/929856104043773963)

### DIFF
--- a/Obsidian/Client.cs
+++ b/Obsidian/Client.cs
@@ -195,12 +195,7 @@ public class Client : IDisposable
                             var loginStart = LoginStart.Deserialize(data);
 
                             string username = config.MulitplayerDebugMode ? $"Player{Globals.Random.Next(1, 999)}" : loginStart.Username;
-                            
-                            if (config.WhitelistEnabled && !config.WhitelistedNicknames.Contains(username))
-                            {
-                                await this.DisconnectAsync("You are not whitelisted on this server\nContact server administrator");
-                                return;
-                            }
+
 
                             this.Logger.LogDebug($"Received login request from user {loginStart.Username}");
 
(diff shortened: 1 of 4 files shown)


Obsidian/Client.cs
```diff
@@ -195,6 +195,12 @@ public async Task StartConnectionAsync()
                             var loginStart = LoginStart.Deserialize(data);
 
                             string username = config.MulitplayerDebugMode ? $"Player{Globals.Random.Next(1, 999)}" : loginStart.Username;
+                            
+                            if (config.WhitelistEnabled && !config.WhitelistedNicknames.Contains(username))
```
Naamloos: This only checks for username. PR is fine otherwise, but would you mind adding an additional check for UUID? UUIDs are generated _after_ the online mode check.

In offline mode you'd check for nickname, yeah, but when in online mode you'd check for UUID. Using the same array is fine.

Line 205 fetches a user from Mojang's API. This response (`var user`) should contain a field named `id`.

It also still needs to convert usernames to uuids before launching the server, but it's fine if you choose to not do that for this PR yet.

Naamloos (changes requested):

Obsidian/Client.cs
roxxel: > This only checks for username. PR is fine otherwise, but would you mind adding an additional check for UUID? UUIDs are generated _after_ the online mode check.
> 
> In offline mode you'd check for nickname, yeah, but when in online mode you'd check for UUID. Using the same array is fine.
> 
> Line 205 fetches a user from Mojang's API. This response (`var user`) should contain a field named `id`.
> 
> It also still needs to convert usernames to uuids before launching the server, but it's fine if you choose to not do that for this PR yet.

Fixed I guess, I can't check if online mode whitelist works 'cause my MC account was stolen

Naamloos (approved):
LGTM

PR #229, opened 2022-01-10
Fixed picked up item's inventory slot

Czompi:


Obsidian.API/Containers/Container.cs
```diff
@@ -30,27 +30,12 @@ public override int AddItem(ItemStack item)
...
-                    invItem.Count += item.Count;
-
-                    return i;
-                }
-
-                if (invItem == null)
-                {
-                    this.items[i] = item;
+            int player_hotbar_start = 36, player_hotbar_end = player_hotbar_start + 8;
+            int player_main_inventory_start = 9, player_main_inventory_end = player_hotbar_start - 1;
```
Naamloos: Generally constant values are defined in the beginning of a class, in all caps. e.g. const int PLAYER_HOTBAR_START, PLAYER_HOTBAR_END

Naamloos (changes requested):

Obsidian.API/Containers/Container.cs
Naamloos: that aside looks great

Seb-stian: In C#, constants are **always** in PascalCase.

Naamloos: oops. yet still they don't belong inline

Obsidian.API/Containers/Container.cs
```diff
@@ -2,6 +2,11 @@
 
 public sealed class Container : BaseContainer, IBlockEntity
 {
+    private const int PLAYER_HOTBAR_START = 36;
+    private const int PLAYER_HOTBAR_END = PLAYER_HOTBAR_START + 8;
+    private const int PLAYER_MAIN_INVENTORY_START = 9;
+    private const int PLAYER_MAIN_INVENTORY_END = PLAYER_HOTBAR_START - 1;
+
```
Czompi: I fixed the naming here

Obsidian.API/Containers/Container.cs
Czompi: > In C#, constants are **always** in PascalCase.

I'll change it.

Czompi: Changed it.

Naamloos (approved):

PR #232, opened 2022-01-19
Overload concat operator for more fluent API access of Chat Messages

DasDarki:
I've overloaded the concat/addition (+) operator in order to generate a more fluent api when creating chat messages.

Here an example:
```csharp
ChatMessage msg = $"{ChatColor.Gray}Welcome {ChatColor.Yellow}you{ChatColor.Gray}! " + ChatMessage.Click("CLICK HERE", EClickAction.RunCommand, "say https://google.com/");
```
I've also added some XML documentation and changed the replace function for reformatting formatting codes using ands (&) to a RegEx based algoritm to still allow and signs in a message.

### DIFF
--- a/Obsidian.API/_Types/ChatMessage.cs
+++ b/Obsidian.API/_Types/ChatMessage.cs
@@ -46,47 +46,65 @@ public class ChatMessage
     /// <summary>
     /// Adds the right <see cref="ChatMessage"/> to the <see cref="Extra"/> of the left <see cref="ChatMessage"/>.
     /// </summary>
-    /// <param name="a">The left chat message on which the right one gets appended</param>
-    /// <param name="b">The right chat message which will be appended</param>
-    /// <returns>The modified chat message</returns>
+    /// <param name="a">The left chat message on which the right one gets appended.</param>
+    /// <param name="b">The right chat message which will be appended.</param>
+    /// <returns>The modified chat message.</returns>
     public static ChatMessage operator +(ChatMessage a, ChatMessage b) => a.AddExtra(b);
 
     /// <summary>
     /// Adds the given <see cref="ChatColor"/> to the text of the given <see cref="ChatMessage"/>.
     /// </summary>
-    /// <param name="a">The message on which the chat color gets appended</param>
-    /// <param name="b">The chat color which will be appended</param>
-    /// <returns>The modified chat message</returns>
+    /// <param name="a">The message on which the chat color gets appended.</param>
+    /// <param name="b">The chat color which will be appended.</param>
+    /// <returns>The modified chat message.</returns>
     public static ChatMessage operator +(ChatMessage a, ChatColor b) => a.AppendColor(b);
 
     /// <summary>
     /// Creates a new <see cref="ChatMessage"/> object with plain text.
     /// </summary>
-    /// <param name="text">The text of the <see cref="ChatMessage"/></param>
-    /// <param name="reformat">Whether to reformat the text via <see cref="ReformatAmpersandPrefixes"/> or not</param>
-    /// <returns>The created <see cref="ChatMessage"/> object</returns>
-    public static ChatMessage Simple(string text, bool reformat = true) => new() { Text = reformat ? ReformatAmpersandPrefixes(text) : text };
+    /// <param name="text">The text of the <see cref="ChatMessage"/>.</param>
+    /// <returns>The created <see cref="ChatMessage"/> object.</returns>
+    public static ChatMessage Simple(string text) => new() { Text = text };
+
+    /// <summary>
+    /// Creates a new <see cref="ChatMessage"/> object with plain text. The text will be reformatted by using
+    /// the <see cref="ReformatAmpersandPrefixes"/> method.
+    /// </summary>
+    /// <param name="text">The text of the <see cref="ChatMessage"/>.</param>
+    /// <returns>The created <see cref="ChatMessage"/> object.</returns>
+    public static ChatMessage SimpleBukkit(string text) => new() { Text = ReformatAmpersandPrefixes(text) };
     
     /// <summary>
     /// Creates a new <see cref="ChatMessage"/> object with plain text.
     /// </summary>
-    /// <param name="text">The text of the <see cref="ChatMessage"/></param>
-    /// <param name="color">The <see cref="ChatColor"/> of the <see cref="ChatMessage"/></param>
-    /// <param name="reformat">Whether to reformat the text via <see cref="ReformatAmpersandPrefixes"/> or not</param>
-    /// <returns>The created <see cref="ChatMessage"/> object</returns>
-    public static ChatMessage Simple(string text, ChatColor color, bool reformat = true) => new()
+    /// <param name="text">The text of the <see cref="ChatMessage"/>.</param>
+    /// <param name="color">The <see cref="ChatColor"/> of the <see cref="ChatMessage"/>.</param>
+    /// <returns>The created <see cref="ChatMessage"/> object.</returns>
+    public static ChatMessage Simple(string text, ChatColor color) => new()
     {
-        Text = $"{color}{(reformat ? ReformatAmpersandPrefixes(text) : text)}"
+        Text = $"{color}{text}"
+    };
+    
+    /// <summary>
+    /// Creates a new <see cref="ChatMessage"/> object with plain text. The text will be reformatted by using
+    /// the <see cref="ReformatAmpersandPrefixes"/> method.
+    /// </summary>
+    /// <param name="text">The text of the <see cref="ChatMessage"/>.</param>
+    /// <param name="color">The <see cref="ChatColor"/> of the <see cref="ChatMessage"/>.</param>
+    /// <returns>The created <see cref="ChatMessage"/> object.</returns>
+    public static ChatMessage SimpleBukkit(string text, ChatColor color) => new()
+    {
+        Text = $"{color}{ReformatAmpersandPrefixes(text)}"
     };
 
     /// <summary>
     /// Appends an underlying <see cref="ClickEvent"/> (<see cref="ClickComponent"/>) to the given <see cref="ChatMessage"/>.
     /// </summary>
-    /// <param name="message">The message which will hold the <see cref="ClickComponent"/></param>
-    /// <param name="action">The action which will be executed when clicking</param>
-    /// <param name="value">The value which will be executed with the action</param>
-    /// <param name="translate">The translate value</param>
-    /// <returns>The given <see cref="ChatMessage"/></returns>
+    /// <param name="message">The message which will hold the <see cref="ClickComponent"/>.</param>
+    /// <param name="action">The action which will be executed when clicking.</param>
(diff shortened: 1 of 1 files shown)


Obsidian.API/_Types/ChatMessage.cs
```diff
@@ -41,11 +44,91 @@ public IEnumerable<ChatMessage> GetExtras()
...
+    /// <param name="translate">The translate value</param>
+    /// <returns>The given <see cref="ChatMessage"/></returns>
+    public static ChatMessage Hover(ChatMessage message, EHoverAction action, object contents, string translate = "")
+    {
+        message.HoverEvent = new HoverComponent(action, contents, translate);
+        return message;
+    }
+
+    /// <summary>
+    /// Converts all formatting codes which are using '&' to there respective '§' formatting code.
```
Seb-stian: using '&' to **their** respective '§'

Obsidian.API/_Types/ChatMessage.cs
```diff
@@ -1,10 +1,13 @@
 ﻿using System.Text.Json;
 using System.Text.Json.Serialization;
+using System.Text.RegularExpressions;
 
 namespace Obsidian.API;
 
 public class ChatMessage
 {
+    private const string AndFormattingCodesPattern = "&(?<code>[a-g0-9k-or])";
```
Seb-stian: Better name might be `AmpersandFormattingCodesPattern`.

Obsidian.API/_Types/ChatMessage.cs
```diff
@@ -41,11 +44,91 @@ public IEnumerable<ChatMessage> GetExtras()
         }
     }
 
+    public static implicit operator ChatMessage(string text) => Simple(text);
+
+    /// <summary>
+    /// Adds the the right <see cref="ChatMessage"/> to the <see cref="Extra"/> of the left <see cref="ChatMessage"/>.
```
Seb-stian: >the the

Obsidian.API/_Types/ChatMessage.cs
```diff
@@ -41,11 +44,91 @@ public IEnumerable<ChatMessage> GetExtras()
...
+    /// <param name="b">The right chat message which will be appended</param>
+    /// <returns>The current chat message. Equally to the left chat message</returns>
+    public static ChatMessage operator +(ChatMessage a, ChatMessage b) => a.AddExtra(b);
+
+    /// <summary>
+    /// Adds the given <see cref="ChatColor"/> to the text of the given <see cref="ChatMessage"/>.
+    /// </summary>
+    /// <param name="a">The message on which the chat color gets appended</param>
+    /// <param name="b">The chat color which will be appended</param>
+    /// <returns>The current chat message. Equally to the given chat message</returns>
```
Seb-stian: The second sentence has confusing wording. I assume you meant to express that it returns the original `ChatMessage` with a modification?

Naamloos:
Welcome! I like this idea. Let me review your code real quick.

Naamloos:
@Seb-stian has a review coming up :p

Obsidian.API/_Types/ChatMessage.cs
```diff
@@ -41,11 +44,91 @@ public IEnumerable<ChatMessage> GetExtras()
...
+        foreach (Match match in matches)
+        {
+            var code = match.Groups["code"].Value;
+            var oldValue = match.Value;
+
+            text = ReplaceFirst(text, oldValue, "§" + code);
+        }
+
+        return text;
+    }
```
Seb-stian: If I understand the code correctly, you iterate over occurences of '§x' and replace them one-by-one. That does not sound ideal. Consider doing something like this instead:
```cs
public static string ReformatFormattingPrefixes(string text)
{
    return string.Create(text.Length, text, (span, text) =>
    {
        for (int i = 0; i < span.Length; i++)
        {
            char c = text[i];
            span[i] = c;

            if (c == '&' && i + 1 < text.Length)
            {
                c = text[i + 1];
                if ((c >= '0' && c <= '9') || (c >= 'a' && c <= 'e') || (c >= 'k' && c <= 'o') || c == 'r')
                {
                    span[i] = '§';
                }
            }
        }
    });
}
```
This would of course remove the need for the regex constant and `ReplaceFirst`.

Also, `ReformatFormattingPrefixes` is a bit confusing (to me, at least). Perhaps `ReformatBukkitPrefixes`/`ReformatAmpersandPrefixes`/`ReplaceAmpersandColors` or something alike would work better?

Obsidian.API/_Types/ChatMessage.cs
```diff
@@ -41,11 +44,91 @@ public IEnumerable<ChatMessage> GetExtras()
...
+    /// <returns>The current chat message. Equally to the given chat message</returns>
+    public static ChatMessage operator +(ChatMessage a, ChatColor b) => a.AppendColor(b);
+
+    /// <summary>
+    /// Creates a new <see cref="ChatMessage"/> object with plain text.
+    /// </summary>
+    /// <param name="text">The text of the <see cref="ChatMessage"/></param>
+    /// <param name="reformat">Whether to reformat the text via <see cref="ReformatFormattingPrefixes"/> or not</param>
+    /// <returns>The created <see cref="ChatMessage"/> object</returns>
+    public static ChatMessage Simple(string text, bool reformat = true) => new() { Text = reformat ? ReformatFormattingPrefixes(text) : text };
```
Seb-stian: I'm not completely sold on this idea. I would like some input from @Naamloos / @Tides here as well. What I don't like about this is how it's going to be used. Look at the following examples:
```cs
var msg = ChatMessage.Simple("&1 Hello World! §r"); // We will just have to remember that both & and § are supported by default, cool.
```
```cs
var msg = ChatMessage.Simple("&1 Hello World! §r", true); // What does the `true` do here? This is going to be confusing.
```
```cs
var msg = ChatMessage.Simple("&1 Hello World! §r", reformat: true); // I guess this helps a little, but at this point I'm thinking that we should just have a separate method for this.
```

Obsidian.API/_Types/ChatMessage.cs
```diff
@@ -41,11 +44,91 @@ public IEnumerable<ChatMessage> GetExtras()
...
+
+    /// <summary>
+    /// Appends an underlying <see cref="ClickEvent"/> (<see cref="ClickComponent"/>) to the given <see cref="ChatMessage"/>.
+    /// </summary>
+    /// <param name="message">The message which will hold the <see cref="ClickComponent"/></param>
+    /// <param name="action">The action which will be executed when clicking</param>
+    /// <param name="value">The value which will be executed with the action</param>
+    /// <param name="translate">The translate value</param>
+    /// <returns>The given <see cref="ChatMessage"/></returns>
+    public static ChatMessage Click(ChatMessage message, EClickAction action, string value, string translate = "")
```
Seb-stian, line 108: I'm fine with this, but I was thinking that having this be an instance method called `WithClick` would be more... conventional? What do you think @Naamloos / @Tides ?

Seb-stian (changes requested):
You seem to end some of your comments with a dot, but not all of them.

I like the idea of using operators for easy concatenation.

DasDarki:
> You seem to end some of your comments with a dot, but not all of them.
> 
> I like the idea of using operators for easy concatenation.

I only use dots in the summary not in the parameter or return description.

Obsidian.API/_Types/ChatMessage.cs
Naamloos: I agree on the separate method. ChatMessage.Simple for simple chat messages, and perhaps ChatMessage.Parsed or ChatMessage.ColorFormatted for the `&` notation? (could use better naming)

Obsidian.API/_Types/ChatMessage.cs
Naamloos: I think this is fine. Would work great with the example given in the original PR message :)

Obsidian.API/_Types/ChatMessage.cs
DasDarki: is obsolete because of the other solution down below

Obsidian.API/_Types/ChatMessage.cs
Seb-stian: Well, both are 'ColorFormatted', but one uses 'Bukkit Color Codes', so that should be reflected in the name I suppose.

Obsidian.API/_Types/ChatMessage.cs
DasDarki: changed

Obsidian.API/_Types/ChatMessage.cs
DasDarki: changed

Obsidian.API/_Types/ChatMessage.cs
DasDarki: changed it

Obsidian.API/_Types/ChatMessage.cs
DasDarki: changed it

Seb-stian:
> I only use dots in the summary not in the parameter or return description.

As I've always seen it and as I've just verified in the [specs](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/documentation-comments#d5-an-example), dots should be used in all XML comments (as they are all sentences).

DasDarki:
> > I only use dots in the summary not in the parameter or return description.
> 
> As I've always seen it and as I've just verified in the [specs](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/documentation-comments#d5-an-example), dots should be used in all XML comments (as they are all sentences).

Alright. Didn't knew that :)

Obsidian.API/_Types/ChatMessage.cs
Seb-stian: @Naamloos , how about:
```cs
public static ChatMessage Simple(string text) => new() { Text = text };
public static ChatMessage SimpleBukkit(string text) => new() { Text = ReformatAmpersandPrefixes(text) };
public static ChatMessage Simple(string text, ChatColor color) => new() { Text = $"{color}{text}" };
public static ChatMessage SimpleBukkit(string text, ChatColor color) => new() { Text = $"{color}{ReformatAmpersandPrefixes(text)}" };
```

Naamloos: perhaps, though you do know you can reuse methods right? :p looks good

Obsidian.API/_Types/ChatMessage.cs
DasDarki: it implemented

Obsidian.API/_Types/ChatMessage.cs
DasDarki: is implemented

Seb-stian: Update: we have an ongoing discussion on the naming still (in the public discord server). This is a tough one 😓

Seb-stian (approved):
LGTM

PR #233, opened 2022-01-21
Adding Player API Methods

DasDarki:
I've added 3 player API methods. 3 PRs for these methods would be somewhat small so I've combined them.

- Action Bar
- Particles
- Potion Effects

### DIFF
--- a/Obsidian.API/_Interfaces/ILiving.cs
+++ b/Obsidian.API/_Interfaces/ILiving.cs
@@ -22,7 +22,7 @@ public interface ILiving : IEntity
     /// <summary>
     /// The dictionary containing all active <see cref="PotionEffect"/> with their respective <see cref="PotionEffectData"/>.
     /// </summary>
-    public IDictionary<PotionEffect, PotionEffectData> ActivePotionEffects { get; }
+    public IReadOnlyDictionary<PotionEffect, PotionEffectData> ActivePotionEffects { get; }
 
     /// <summary>
     /// Whether the entity has the given <see cref="PotionEffect"/> or not.
--- a/Obsidian.API/_Interfaces/ILiving.cs
+++ b/Obsidian.API/_Interfaces/ILiving.cs
@@ -41,27 +41,12 @@ public interface ILiving : IEntity
     /// </summary>
     /// <param name="potion">The potion effect to be added.</param>
     /// <param name="duration">The duration of the potion in ticks.</param>
-    public Task AddPotionEffectAsync(PotionEffect potion, int duration);
-
-    /// <summary>
-    /// Adds the given <see cref="PotionEffect"/> to the entity.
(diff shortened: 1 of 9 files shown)


Naamloos (approved):
LGTM

Obsidian.API/_Interfaces/ILiving.cs
```diff
@@ -16,4 +16,54 @@ public interface ILiving : IEntity
     public int AbsorbtionAmount { get; set; }
 
     public Vector BedBlockPosition { get; set; }
+    
+    /// <summary>
+    /// The list containing all currently active <see cref="PotionEffect"/>s.
+    /// </summary>
+    public IReadOnlyList<PotionEffect> ActivePotionEffects { get; }
```
Tides: This should be a dictionary that holds more relevant information i.e a class that has duration, amplifiers, flags and ofc with the effect as its key ^ ^ instead of having what you did below with "PotionEffectDurations"

Tides (changes requested):
Everything else LGTM 👍🏼

Obsidian.API/_Interfaces/ILiving.cs
DasDarki: @Tides changed it

Tides: Looks good thanks!

Tides (approved):

Obsidian/Entities/Living.cs
```diff
@@ -15,6 +17,75 @@ public class Living : Entity, ILiving
...
+    }
+    
+    public Task AddPotionEffectAsync(PotionEffect potion, int duration) => AddPotionEffectAsync(potion, duration, 0);
+
+    public Task AddPotionEffectAsync(PotionEffect potion, int duration, byte amplifier) =>
+        AddPotionEffectAsync(potion, duration, amplifier, true, true, false);
+
+    public async Task AddPotionEffectAsync(PotionEffect potion, int duration, byte amplifier, bool showParticles,
+        bool showIcon, bool isAmbient)
+    {
```
Craftplacer: This could've made use of [C#'s optional parameters](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/named-and-optional-arguments#optional-arguments) instead of creating method overloads.

Craftplacer (changes requested):
LGTM, but I'd like to see that nit addressed.

Obsidian/Entities/Living.cs
Naamloos: +1 on that one.

DasDarki:
> LGTM, but I'd like to see that nit addressed.

Is fixed

Craftplacer (approved):

PR #234, opened 2022-01-21
Simplify DataArray

Seb-stian:
- No more magic numbers
- No more naught-saying variable names
- Reduced complexity
- Unchanged performance


Tides (approved):
I like what I see 🙏🏼

PR #240, opened 2022-01-27
Cleanup MainCommandModule.cs

Tides:
🙏🏼


Craftplacer (commented):
What happened with the permission commands that you removed?

Tides:
> What happened with the permission commands that you removed?

The permission system is unused/not working and needs a rework ^ ^ So I removed the commands as those are most likely to change as well

Craftplacer:
> > What happened with the permission commands that you removed?
> 
> The permission system is unused/not working and needs a rework ^ ^ So I removed the commands as those are most likely to change as well

Aight

Craftplacer (approved):

PR #241, opened 2022-01-31
Moderation Commands

punkroczombie:
Created Ban/Unban/Kick commands for future moderation

### DIFF
--- a/Obsidian.API/_Interfaces/IBan.cs
+++ b/Obsidian.API/_Interfaces/IBan.cs
@@ -1,0 +1,9 @@
+namespace Obsidian.API;
+
+public interface IBan
+{
+    public string Username { get; set; }
+    public Guid Uuid { get; set; }
+    public int Duration { get; set; }
+    public DateTime TimeStamp { get; set; }
+}
--- a/Obsidian.API/_Interfaces/IBan.cs
+++ b/Obsidian.API/_Interfaces/IBan.cs
@@ -4,6 +4,6 @@ public interface IBan
 {
     public string Username { get; set; }
     public Guid Uuid { get; set; }
-    public int Duration { get; set; }
+    public TimeSpan? Duration { get; set; }
     public DateTime TimeStamp { get; set; }
 }
--- a/Obsidian.API/_Interfaces/IBanList.cs
+++ b/Obsidian.API/_Interfaces/IBanList.cs
@@ -1,0 +1,13 @@
+namespace Obsidian.API;
+
+public interface IBanList
+{
+    public void AddBan(IPlayer player, int duration);
+    public void AddBan(string username, int duration);
+    public void RemoveBan(IPlayer player);
+    public void RemoveBan(string username);
+    public IBan GetBan(IPlayer player);
+    public IBan GetBan(string username);
+    public bool IsBanned(IPlayer player);
+    public bool IsBanned(string username);
+}
--- a/Obsidian.API/_Interfaces/IBanList.cs
+++ b/Obsidian.API/_Interfaces/IBanList.cs
@@ -2,9 +2,9 @@ namespace Obsidian.API;
 
 public interface IBanList
 {
-    public List<IBan> Bans { get; set;}
-    public void AddBan(IPlayer player, int? duration);
-    public void AddBan(string username, int? duration);
+    public IList<IBan> Bans { get; }
+    public void AddBan(IPlayer player, TimeSpan? duration);
+    public void AddBan(string username, TimeSpan? duration);
     public void RemoveBan(IPlayer player);
     public void RemoveBan(string username);
     public IBan GetBan(IPlayer player);
--- a/Obsidian.API/_Interfaces/IBanList.cs
+++ b/Obsidian.API/_Interfaces/IBanList.cs
@@ -2,8 +2,9 @@ namespace Obsidian.API;
 
 public interface IBanList
 {
-    public void AddBan(IPlayer player, int duration);
-    public void AddBan(string username, int duration);
+    public List<IBan> Bans { get; set;}
+    public void AddBan(IPlayer player, int? duration);
+    public void AddBan(string username, int? duration);
     public void RemoveBan(IPlayer player);
     public void RemoveBan(string username);
     public IBan GetBan(IPlayer player);
--- a/Obsidian/Client.cs
+++ b/Obsidian/Client.cs
@@ -179,7 +179,7 @@ public class Client : IDisposable
                             if (this.Server.Bans.IsBanned(this.Player))
                             {
                                 IBan ban = this.Server.Bans.GetBan(this.Player);
-                                if (DateTime.Now.CompareTo(ban.TimeStamp.AddDays(ban.Duration)) > 0)
+                                if (DateTime.Now.CompareTo(ban.Duration) <= 0)
                                 {
                                     this.Server.Bans.RemoveBan(this.Player);
                                 }
--- a/Obsidian/Client.cs
+++ b/Obsidian/Client.cs
@@ -176,16 +176,6 @@ public class Client : IDisposable
 
                         if(nextState == ClientState.Login)
                         {
-                            if ((int)handshake.Version > (int)Server.Protocol)
-                            {
-                                await this.DisconnectAsync($"Outdated server! I'm still on {Server.Protocol.GetDescription()}.");
-                            }
-
-                            if ((int)handshake.Version < (int)Server.Protocol)
-                            {
-                                await this.DisconnectAsync($"Outdated client! Please use {Server.Protocol.GetDescription()}.");
-                            }
-
                             if (this.Server.Bans.IsBanned(this.Player))
                             {
                                 IBan ban = this.Server.Bans.GetBan(this.Player);
--- a/Obsidian/Client.cs
+++ b/Obsidian/Client.cs
@@ -174,8 +174,36 @@ public class Client : IDisposable
                             await this.DisconnectAsync("you seem suspicious");
                         }
 
-                        this.State = nextState;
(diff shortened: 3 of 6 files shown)


Obsidian/Utilities/BanList.cs
```diff
@@ -0,0 +1,95 @@
...
+        return this.bans.FirstOrDefault(ban => ban.Username == player.Username);
+    }
+
+    public IBan GetBan(string username)
+    {
+        return this.bans.FirstOrDefault(ban => ban.Username == username);
+    }
+
+    public bool IsBanned(IPlayer player) => this.bans.Any(x => x.Username == player.Username || player.Uuid == x.Uuid);
+    public bool IsBanned(string username) => this.bans.Any(x => x.Username == username || x.Uuid == Guid.Parse(username));
```
DasDarki, line 81: I'm not exactly sure but the `Guid.Parse(username)` could throw an exception if the method gets called and the username parameter is indeed a username but no user was found. The Guid.Parse would then be called and would throw an exception. It would be better if you use `Guid.TryParse`.

Naamloos (dismissed):
LGTM honestly

Naamloos:
Thanks for the contribution! :)

Obsidian/Commands/MainCommandModule.cs
```diff
@@ -331,6 +331,38 @@ public Task StopAsync(CommandContext ctx)
...
 
+    [Command("ban")]
+    [CommandInfo("Bans a user.", "/ban <user> <duration>")]
+    [RequirePermission(permissions: "obsidian.ban")]
+    public async Task BanAsync(CommandContext ctx, string username, int duration)
+    {
+        var server = (Server)ctx.Server;
+        var player = server.GetPlayer(username);
+        server.Bans.AddBan(player, duration);
+        await player.KickAsync($"{username} was banned by an Operator");
```
roxxel: It would be better to specify the operator's nickname than just the operator (or if it was console/rcon then "Console")
same goes to other commands

roxxel: There's IsPlayer property in CommandContext, u can use it to check whether command was executed by player or not

punkroczombie: not a bad idea

Obsidian.API/_Interfaces/IBanList.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+    public void AddBan(IPlayer player, int duration);
+    public void AddBan(string username, int duration);
+    public void RemoveBan(IPlayer player);
+    public void RemoveBan(string username);
+    public IBan GetBan(IPlayer player);
+    public IBan GetBan(string username);
+    public bool IsBanned(IPlayer player);
+    public bool IsBanned(string username);
+}
```
Craftplacer: Would like if you could iterate over it, like a list. (Maybe make it inherit `IList<IBan>`?)

Obsidian.API/_Interfaces/IBan.cs
```diff
@@ -0,0 +1,9 @@
+namespace Obsidian.API;
+
+public interface IBan
+{
+    public string Username { get; set; }
+    public Guid Uuid { get; set; }
+    public int Duration { get; set; }
```
Craftplacer: Duration as int doesn't feel great. A `TimeSpan` would've been better IMO.

Obsidian/Commands/MainCommandModule.cs
```diff
@@ -331,6 +331,62 @@ public Task StopAsync(CommandContext ctx)
...
+        if (ctx.IsPlayer)
+        {
+            await player.KickAsync($"{username} was banned by {ctx.Player.Username}");
+        }
+        else
+        {
+            await player.KickAsync($"{username} was banned by Console");
+        }
+    }
+
```
Craftplacer, line 345: Why is duration mandatory? It should be possible to permanently ban players. Consider making Duration nullable.

Obsidian/Commands/MainCommandModule.cs
```diff
@@ -331,6 +331,62 @@ public Task StopAsync(CommandContext ctx)
...
+        var player = server.Players.FirstOrDefault(player => player.Username == username);
+        
+        if (ctx.IsPlayer)
+        {
+            await player.KickAsync($"{username} was banned by {ctx.Player.Username}");
+        }
+        else
+        {
+            await player.KickAsync($"{username} was banned by Console");
+        }
```
Craftplacer: Copy pasta fail, the disconnect message is wrong.

Craftplacer (changes requested):
Already a good start but many opportunities to improve.

Obsidian.API/_Interfaces/IBanList.cs
punkroczombie: would List<IBan> work?

Obsidian/Commands/MainCommandModule.cs
roxxel: Also, I suggest using ternary operator
```cs
await player.KickAsync($"{username} was banned by {(ctx.IsPlayer ? ctx.Player.Username : "Console")}");
```
Not sure if parentheses is required, check if it compiles without them

Obsidian/Commands/MainCommandModule.cs
Naamloos: +1

Obsidian.API/_Interfaces/IBanList.cs
Craftplacer: `IList<IBan>` would be better, since this is already an interface.

Obsidian.API/_Interfaces/IBanList.cs
```diff
@@ -0,0 +1,14 @@
+namespace Obsidian.API;
+
+public interface IBanList
+{
+    public List<IBan> Bans { get; set;}
+    public void AddBan(IPlayer player, int? duration);
+    public void AddBan(string username, int? duration);
```
Craftplacer: Duration is still an `int` internally. This is an issue because you, as a developer, don't know what unit this int is supposed to represent. Also, by looking at the code, this only accommodates days and not smaller units like hours, minutes and seconds -- which may be preferred by admins or something.

Obsidian/Client.cs
```diff
@@ -176,6 +176,19 @@ public async Task StartConnectionAsync()
 
                         if(nextState == ClientState.Login)
                         {
+                            if (this.Server.Bans.IsBanned(this.Player))
+                            {
+                                IBan ban = this.Server.Bans.GetBan(this.Player);
+                                if (DateTime.Now.CompareTo(ban.Duration) <= 0)
```
Craftplacer, line 182: This code doesn't seem right, is an even `int` expected?
https://docs.microsoft.com/en-us/dotnet/api/system.datetime.compareto#system-datetime-compareto(system-object)

Obsidian/Commands/MainCommandModule.cs
```diff
@@ -331,6 +331,41 @@ public Task StopAsync(CommandContext ctx)
         return Task.CompletedTask;
     }
 
+    [Command("ban")]
+    [CommandInfo("Bans a user.", "/ban <user> ?<duration>")]
```
Craftplacer: `[]` are the correct brackets for indicated optional parameters.

Obsidian/Utilities/BanList.cs
```diff
@@ -0,0 +1,110 @@
...
+    {
+        var fi = new FileInfo(this.Path);
+
+        if (fi.Exists)
+        {
+            using var fs = fi.OpenRead();
+            this.Bans = await fs.FromJsonAsync<List<IBan>>();
+            foreach (Ban ban in Bans)
+            {
+                if (DateTime.Now.CompareTo(ban.Duration) <= 0)
```
Craftplacer, line 29: Same as before.

Craftplacer (changes requested):

Obsidian.API/_Interfaces/IBanList.cs
punkroczombie: got it

Obsidian/Client.cs
punkroczombie: it worked when I tested it is there a better way to check?

Craftplacer: Check an example I wrote: https://rextester.com/YNVXS46468

Craftplacer: The example is pretty basic since you only need to calculate the time (stamp) when the player would be considered unbanned, and then compare whether the current time exceeds that time stamp.

punkroczombie: Thanks ill try that out

Obsidian/Utilities/BanList.cs
```diff
@@ -0,0 +1,111 @@
...
+    public void AddBan(IPlayer player, TimeSpan? duration)
+    {
+        if (duration is null)
+        {
+            bans.Add(new Ban { Username = player.Username, Uuid = player.Uuid, Duration = null, TimeStamp = DateTime.Now });
+        }
+        else
+        {
+            bans.Add(new Ban { Username = player.Username, Uuid = player.Uuid, Duration = duration, TimeStamp = DateTime.Now });
+        }
```
Seb-stian, line 55: This condition seems redundant.

Obsidian/Utilities/BanList.cs
```diff
@@ -0,0 +1,111 @@
...
+    public void AddBan(string username, TimeSpan? duration)
+    {
+        if (duration is null)
+        {
+            this.bans.Add(new Ban { Username = username, Uuid = Guid.Empty, Duration = null, TimeStamp = DateTime.Now });
+        }
+        else
+        {
+            this.bans.Add(new Ban { Username = username, Uuid = Guid.Empty, Duration = duration, TimeStamp = DateTime.Now });
+        }
```
Seb-stian, line 69: This condition seems redundant.

XanderStoffels:
What will happen to this PR?

Craftplacer:
> What will happen to this PR?

Either some other pull request will make this one redundant, or it won't be updated in which case we might just close it.

Naamloos:
It has been open for over a year so since requested changes have not been taken into account I will be closing this PR.

Obviously, feel free to reopen the pull request with requested changes and conflicts resolved.

PR #242, opened 2022-02-14
Fix slow dictionary usage

Seb-stian:
It all comes down to
```diff
- return this.Regions.SingleOrDefault(x => x.Key == value).Value;
+ return Regions.TryGetValue(value, out Region? region) ? region : null;
```
as it turns out, iterating over a whole dictionary is a little bit slower than accessing elements by a key.

With this change, generation time (in debug mode) went down from 4:18 to 2:46!


Jonpro03 (approved):
LGTM

Naamloos (approved):
MTGL

PR #243, opened 2022-02-14
Refactor Client class

Seb-stian:
Notable changes:
- fixed some (not all) nullable warnings
- uses Socket directly, instead of TcpClient
- fixes a bug where client can't connect again after disconnecting

Since there are a lot of small changes, you might prefer to review the changed file directly [here](https://github.com/ObsidianMC/Obsidian/blob/client-refactor/Obsidian/Client.cs).

### DIFF
--- a/Obsidian/Entities/Player.cs
+++ b/Obsidian/Entities/Player.cs
@@ -198,7 +198,7 @@ public class Player : Living, IPlayer
     {
         this.LastPosition = this.Position;
         this.Position = pos;
-        await this.client.UpdateChunksAsync(true);
+        await this.UpdateChunksAsync(true);
 
         var tid = Globals.Random.Next(0, 999);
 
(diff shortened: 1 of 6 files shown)


Obsidian/Entities/Player.cs
```diff
@@ -880,6 +880,57 @@ private async Task PickupNearbyItemsAsync(float distance = 0.5f)
         }
     }
 
+    internal async Task UpdateChunksAsync(bool unloadAll = false)
```
Jonpro03, line 883: I would argue that updating chunks is Client and not Player, but I think @Tides agrees with you too so majority rules.

Jonpro03 (approved):
Yeah! This looks good

Obsidian/Entities/Player.cs
Naamloos: fun fact: I personally agree on chunks being a client thing.

Naamloos (commented):
Some discussion needed

Naamloos (approved):

PR #245, opened 2022-02-25
Change player names to be case-insensitive

Seb-stian:
Since minecraft names are case-insensitive, it makes sense to do their comparisons as such.

Another possible design would be something like:
```cs
public bool IsPlayerOnline(string username, StringComparison comparison = StringComparison.InvariantCultureIgnoreCase);
```
or
```cs
public bool IsPlayerOnline(string username, bool caseSensitive = false);
```

### DIFF
--- a/Obsidian/Server.cs
+++ b/Obsidian/Server.cs
@@ -155,13 +155,13 @@ public partial class Server : IServer
     /// </summary>
     /// <param name="username">The username you want to check for.</param>
     /// <returns>True if the player is online.</returns>
-    public bool IsPlayerOnline(string username) => OnlinePlayers.Any(x => x.Value.Username == username);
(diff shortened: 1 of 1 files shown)


Naamloos:
Minecraft's player names are case insensitive. However, realistically you'd want to check for uuid instead.

Seb-stian:
> However, realistically you'd want to check for uuid instead.

Plugins may work with player names if they load them from configuration/save files. The latter should realistically contain case-matching names, the former may not.

But either way, I see this as a reasonable change, whether the API is realistically going to be used or not. Don't you think so? 🧐

Jonpro03 (approved):

Naamloos:
aye chief

PR #246, opened 2022-02-25
Remove plugin services and sandboxing

Seb-stian:
Plugin services came to be for two reasons:
- ease of developing
- safety

Plugin permissions came to be for **security** reasons only.

After research, it was determined that sandboxing in .NET is **not** realistically possible. To do sandboxing with actual safety guarantees, we would have to do it at an OS level, running plugins in separate processes. Such concept may be researched further in the future, but is out of scope for this PR.

Since plugin services were also introduced for convenience, they may make a comeback (and so not all of the service API is removed).


Naamloos (approved):
Nice

PR #250, opened 2022-02-26
Fix subgroup commands

Seb-stian:
```diff
- private void RegisterSubcommands(Type type, PluginContainer plugin, object instance, Command parent = null)
```
`instance` was unused, but the method was being called as:
```cs
RegisterSubcommands(st, plugin, cmd);
```
`cmd` was supposed to be `parent`, but it was used as `instance` instead, causing `parent` to be `null`, causing a command to have no overloads registered. This has lead to the command always causing an argument mismatch error.

Furthermore, command groups had their correct parent type, but subgroups didn't, causing `TargetMethodException` when trying to invoke the command method on a wrong type.


Naamloos (approved):
yes

PR #258, opened 2022-04-30
Use faster VarInt byte count calculation algorithm

Seb-stian:
Benchmarks are looking good ✔️ 
Test are passing ✔️ 

Close #257


Naamloos (approved):
great!

PR #259, opened 2022-05-05
RCON server

### LINKED ISSUE #82: Implement RCON abilities
The current console does not support user input. I don't think we'll ever implement this, because we run multiple servers on one console. A different option is implementing Minecraft's RCON protocol, so we can communicate with one specific server remotely.

https://wiki.vg/RCON

Hacktoberfest edit: welcome! Feel free to tackle this issue if you feel like it. The RCON protocol would have to be it's own class that is able to accept connections and interact with the console in/output from Obsidian. Good luck!

ndelta0:
Working RCON communications (vanilla + extended with added security by using a pre-shared key or Diffie-Hellman exchange).

Will resolve #82 once complete.

It's a draft because of the incomplete command system.

Link to a compatible RCON client using encryption [here, a zip file with source code](https://cdn.discordapp.com/attachments/854801139219300402/971834898539225148/RconClient.zip).


Obsidian.API/_Interfaces/IConfig.cs
```diff
@@ -58,4 +58,35 @@ public interface IConfig
...
+
+    /// <summary>
+    /// Key used in RCON encryption.
+    /// </summary>
+    public string? RconKey { get; set; }
+
+    /// <summary>
+    /// Whether the RCON encryption can happen with a Diffie-Hellman Key Exchange or only PSK.
+    /// </summary>
+    public bool AllowDiffieHellman { get; set; }
```
Craftplacer: Considering encryption is a fragile subject, this should be an enum to allow for future changes.

ndelta0: By enum do you mean only one encryption method or multiple possibilities at once

ndelta0:
~~Broadcasting to OPs still missing~~

Obsidian/Net/Rcon/RconConnection.cs
```diff
@@ -244,6 +244,13 @@ await SendAsync(new RconPacket(currentEncoding)
 
                         try
                         {
+                            logger.LogInformation("Executing '{Prefix}{Command}'", CommandHandler.DefaultPrefix, packet.PayloadText);
```
Craftplacer, line 247: Debug log message in information scope is bad

Obsidian/Net/Rcon/RconConnection.cs
```diff
@@ -244,6 +244,13 @@ await SendAsync(new RconPacket(currentEncoding)
...
+
+                            server.Operators.GetOnlineOperators().ForEach(op =>
+                            {
+                                op.SendMessageAsync($"[RCON: {CommandHandler.DefaultPrefix}{packet.PayloadText}]", MessageType.System);
```
Craftplacer, line 251: We should have a more public/general method to broadcast command execution messages instead of only putting it into RconConnection.cs

Craftplacer (changes requested):

Obsidian/Utilities/Config.cs
```diff
@@ -37,6 +37,15 @@ public class Config : IConfig
...
 
+    public bool EnableRcon { get; set; } = false;
+
+    public string RconPassword { get; set; } = string.Empty;
+
+    public ushort RconPort { get; set; } = 25575;
+
+    public bool BroadcastRconToOps { get; set; } = true;
+    public bool RequireRconEncryption { get; set; } = false;
+
```
Craftplacer: Would consider moving these properties into their own class so the resulting json would look like

```json
{
  "rcon": {
    "enabled": true,
    "password": "xyz",
    "port": 6969,
    "broadcastCommandsToOperators": false,
    "requireEncryption": false 
  }
}
```

Obsidian/Server.cs
```diff
@@ -655,7 +663,7 @@ private async Task LoopAsync()
     /// <summary>
     /// Registers the "obsidian-vanilla" entities and objects.
     /// </summary>
-    /// Might be used for more stuff later so I'll leave this here - tides 
+    /// Might be used for more stuff later so I'll leave this here - tides
```
Craftplacer, line 666: Remove the additional slash here as it's not part of documentation.

Obsidian/Net/Rcon/RconServer.cs
```diff
@@ -0,0 +1,96 @@
...
+                }
+                catch (Exception e) when (e is TaskCanceledException or OperationCanceledException)
+                {
+                    return;
+                }
+        }, token);
+
+        connectionId = 0;
+        listener.Start();
+        logger.LogInformation("Started RCON server");
```
Craftplacer, line 75: Information is the wrong scope for these messages, verbose or debug might be more fitting

Craftplacer (changes requested):

Obsidian/Net/Rcon/RconConnection.cs
ndelta0: It's not for debugging purposes, Notchian server does the same. Also, would you rather know if someone got unauthorized access to the server through RCON?

Obsidian/Net/Rcon/RconConnection.cs
ndelta0: I don't see any other use case for sending something to all operators, also it would've been possible to refactor this at a later point if it's needed.

Obsidian/Utilities/Config.cs
ndelta0: Possible

Obsidian/Server.cs
ndelta0: What?

Obsidian/Net/Rcon/RconServer.cs
ndelta0: It's literally an informational thing, starting an RCON server is information. Yes I'd have put this as debug or verbose if it was called multiple times, but it's literally a one-time thing.

Obsidian/Net/Rcon/RconConnection.cs
Craftplacer: Notchian servers inform an operator about the output from commands executed by other operators

Obsidian/Server.cs
Craftplacer: Tides forgot to remove the third slash from the comment.

Obsidian/Net/Rcon/RconConnection.cs
ndelta0: Sorry to inform you, but RCON also [as stated here](https://minecraft.fandom.com/wiki/Server.properties#Java_Edition_2:~:text=online%20operators.-,broadcast%2Drcon%2Dto%2Dops,Send%20rcon%20console%20command%20outputs%20to%20all%20online%20operators.,-difficulty).

Also sending just the command output is not a sensible option until the rewrite/rework of the command system. RCON just sending `help` would print the help to all OPs.

Jonpro03 (dismissed):
Looks good. This must've been a ton of work! Thank you.
My only gripe is that there is a lot of nesting in RconConnection.cs but I've also seen and written worse.

Cheers!

ndelta0:
> [...] My only gripe is that there is a lot of nesting in RconConnection.cs but I've also seen and written worse.

Unfortunately, that was the most straightforward way of handling authentication and encryption I've came up with. As always, any suggestions are welcome.

Craftplacer (approved):
looks ok i think

Jonpro03 (approved):

Issue #261, opened 2022-05-09
Update to preview versions of .NET


Seb-stian:
Before the first official Obsidian release, we can make use of preview .NET technologies. There is no reason not to, as we don't have any end-customers who may suffer from instability. This will help us to keep up with the newest features as they come out, instead of doing bulk updates yearly.

Currently this means updating to .NET 7 Preview 3

Seb-stian:
@Naamloos Any further suggestions of what should be done after updating?

- use Regex source generators
- fix possible [breaking changes](https://docs.microsoft.com/en-us/dotnet/core/compatibility/7.0)
- ...

Naamloos:
Honestly I'm not too sure lol

But being up to date is always fun

Seb-stian:
No longer relevant

PR #262, opened 2022-05-10
Bump .NET version to 7

Naamloos:
implements #261


Craftplacer:
Is it necessary to explicitly declare the .NET version inside the job?

I am sure .NET lets you build for older versions. We should also leverage environment variables within the job.

Naamloos:
That's something for another day, feel free to open an issue :)

PR #265, opened 2022-05-18
Fix console input blocking server in linux

### LINKED ISSUE #264: Console.ReadKey() blocking the entire server in Linux
While testing the server for the first time on a Linux PC, I found a bug. 

When the console app waits for a keypress, the entire server blocks and doesn't log anything or send any packets to the client until the enter key is pressed (one or multiple times).
image

If no one is against it, I'll make a pull request now with the fix.

<sub>Sorry for my bad English, I'm not a native speaker<sub>

GasInfinity:
Fixes #264

### DIFF
--- a/Obsidian/Utilities/ConsoleIO.cs
+++ b/Obsidian/Utilities/ConsoleIO.cs
@@ -6,6 +6,8 @@
 /// Provide some fancy features such as formatted output, text pasting and tab-completion.
 /// By ORelio - (c) 2012-2018 - Available under the CDDL-1.0 license
 /// source https://github.com/ORelio/Minecraft-Console-Client/blob/master/MinecraftClient/ConsoleIO.cs
+/// 
+/// Modified by GasInfinity -> Fix Linux Console Blocking - 18/05/2022
 /// </summary>
 public static class ConsoleIO
 {
--- a/Obsidian/Utilities/ConsoleIO.cs
+++ b/Obsidian/Utilities/ConsoleIO.cs
@@ -1,4 +1,6 @@
-﻿namespace Obsidian.Utilities;
+﻿using System.Threading;
+
+namespace Obsidian.Utilities;
 
 /// <summary>
 /// Allows simultaneous console input and output without breaking user input
@@ -80,7 +82,12 @@ public static class ConsoleIO
 
         while (k.Key != ConsoleKey.Enter)
         {
-            while(!Console.KeyAvailable) {} // Don't block, just wait in a loop until there's a KeyPress
+            while(!Console.KeyAvailable) 
+            {
+                // This doesn't introduce noticeable input lag. Maybe in the future we'll come with a better solution without any input lag?
+                Thread.Sleep(50);
+            }
+
             k = Console.ReadKey(true);
 
             lock (io_lock)
--- a/Obsidian/Utilities/ConsoleIO.cs
+++ b/Obsidian/Utilities/ConsoleIO.cs
@@ -59,6 +59,7 @@ public static class ConsoleIO
     public static bool BasicIO_NoColor = false;
 
 
+    private static Func<bool, ConsoleKeyInfo> PlatformReadKey;
 
     /// <summary>
     /// Read a line from the standard input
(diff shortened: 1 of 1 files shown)


Seb-stian (changes requested):
Thank you for the PR. I see few problems here. I believe the license on this file requires you to

> include a notice in each of Your Modifications that identifies You as the Contributor of the Modification

Furthermore, although it is (I assume) a working solution, active waiting is not ideal. We can merge it for now (after the licensing issue is solved) and improve it later (let's not forget to open a new issue for it, if we go down that route).

Seb-stian (approved):

GasInfinity:
> Furthermore, although it is (I assume) a working solution, active waiting is not ideal. We can merge it for now (after the licensing issue is solved) and improve it later (let's not forget to open a new issue for it, if we go down that route).

Yes, active waiting is not ideal. I was thinking about doing a ```Task.Delay()``` but that would introduce some input lag.

Edit: I'm very dumb, that isn't an async function

Seb-stian:
Can you confirm that this actually works? Did you test it?

GasInfinity:
Yes, I'm actually connected to my local server now and it works. Now before this gets merged I want to resolve the empty while loop, i think that a simple `Thread.Sleep()` would work but as I've said before that could introduce some input lag.

Craftplacer:
Does the bug appear without this PR? Polling for an available key is an odd solution to me.

Seb-stian:
@Craftplacer 
> Does the bug appear without this PR?

This is a known bug that has been reported by different users before.

GasInfinity:
> Does the bug appear without this PR? Polling for an available key is an odd solution to me.

Yes, without this PR, the server is practically unusable in Linux (at least on my PC).

There are other solutions like enabling BasicIO, but you loose some useful features.

I can't test this PR in Windows right now, but I think it should work fine there.

Craftplacer:
I don't understand why this fix is necessary in the first place (why doesn't ReadKey work under Linux?), is this a framework bug?

GasInfinity:
> I don't understand why this fix is necessary in the first place (why doesn't ReadKey work under Linux?), is this a framework bug?

I think this could be related:
https://github.com/dotnet/runtime/issues/34239

Edit: I'm 90% sure that that's the problem.

Jonpro03:
> > I don't understand why this fix is necessary in the first place (why doesn't ReadKey work under Linux?), is this a framework bug?
> 
> I think this could be related: [dotnet/runtime#34239](https://github.com/dotnet/runtime/issues/34239)
> 
> Edit: I'm 90% sure that that's the problem.

Thank you for finding that. 
It was driving me nuts not knowing why linux was broken.

Jonpro03 (approved):

Naamloos:
Personally I think it's best to ignore console input altogether now that rcon is coming...

Craftplacer:
> > I don't understand why this fix is necessary in the first place (why doesn't ReadKey work under Linux?), is this a framework bug?
> 
> I think this could be related: [dotnet/runtime#34239](https://github.com/dotnet/runtime/issues/34239)
> 
> Edit: I'm 90% sure that that's the problem.

We should redo this then and come up with this design:

Thread 1: Dequeues WriteLine's and handles console I/O
Thread 2, 3, ...: Queue WriteLines to the host console

This design should only apply to the host console I think, unnecessary for other ones

GasInfinity:
> > > I don't understand why this fix is necessary in the first place (why doesn't ReadKey work under Linux?), is this a framework bug?
> > 
> > I think this could be related: [dotnet/runtime#34239](https://github.com/dotnet/runtime/issues/34239)
> > 
> > Edit: I'm 90% sure that that's the problem.
> 
> We should redo this then and come up with this design:
> 
> Thread 1: Dequeues WriteLine's and handles console I/O
> Thread 2, 3, ...: Queue WriteLines to the host console
> 
> This design should only apply to the host console I think, unnecessary for other ones

That should work, the other threads won't block.

But if we want a fancy status line and some other features we would still need to check for `Console.KeyAvailable`, because those WriteLines would need to wait for a key to be written on the Console.

But I think that it would be better to open other PR later to redo that.

Craftplacer:
I would recommend to wrap the while statement with a platform check (this is Linux only apparently) so we can finally merge this fix.

Issue #270, opened 2022-07-29
A list of minor and slightly minor flaws


Qialet:
A list of shortcomings related to the game itself or other problems. 

I think it would be time to make a list in which there will be all the minor errors related to the kernel, so as not to lose anything. 
Bugs were found on this build: https://github.com/ObsidianMC/Obsidian/actions/runs/2754445504 
Some errors may have already been found, so I apologize in advance.

1. **CRITICAL.** After the first start of the kernel, and then switching on 2 times, and logging on to the server, it simply does not load chunks, and the model falls down.
At the same time, all commands are executed.
2. The /gamemode creative command works even when the user does not have /op rights... Although the /stop command does not work.
3. https://github.com/ObsidianMC/Obsidian/issues/282
4. When logging in to the server: (the server does not restart)
  - My spawned underdragon via Spawnentity has disappeared 
  - Things lying on the ground are gone
  - The first slot always becomes the active item

5. If you click on the grass, snow, leana with the right mouse button, 2 blocks will be put instead of one
(the block will be put in the place where the grass was, the second block is put in the direction where the cursor looks when the first block was put)

5. White block of land. If you put a block of land, it will turn white...

6. Blocks can be put up. Put 2 blocks in height and put the grass in the other direction or also up, and it will fly.
Blocks that this shouldn't work with:
  - Grass, flowers.
  - Rails
  - Snow
  - Stairs
7.Blocks that do not connect (at the same time have strange textures)
  - Fences (stone wooden)
  - Iron bars
8. If the grass has a height of 2 blocks, break the upper or lower part, then the 2nd part will not break.
9. If you go to your inventory in the creative, hold down shift, and press the delete items button, then in fact the inventory should be cleared, which does not happen.
10. Leans have the form of a square block, literally. (if you put a block next to them, then leana will take a normal look)
11. You can not write on the Plates
12. **CRITICAL.** If you put an anvil or sand in such a way that it falls from a height, your model will simply stop moving (solved by re-entering the server)
13. If there is less than 1 block of height between the sand or anvil from which they will fall, they will fall faster than usual.
14. Only a completely whole anvil works, you cannot interact with others.
15. Blocks can be placed up even if they are placed under themselves
16. Gravel has no physics. (does not fall)
17. An empty stove, a captille, a melting furnace and a potion maker has animations and sounds when she cooks something. (when re-entering, the light from it disappears)
18. The stove does not accept these wooden things (wood of other types as well)
    1) oak_wood
[...]

PR #278, opened 2022-09-15
Switching game modes now requires OP. Provide user feedback on command failure.

XanderStoffels:
Addresses point 2 in https://github.com/ObsidianMC/Obsidian/issues/270.

While trying to fix this, I encountered a todo and gave it a shot:
`                // A check failed.
                // TODO: Tell user what arg failed?`

Instead of printing the `CommandExecutionCheckException` to console, exceptions inheriting `CommandExecutionCheckException` will now get caught and can provide feedback to the command executor if applicable.


Naamloos (approved):
Honestly I don't see anything wrong with this. Thank you for your first contribution to the project! :)

Craftplacer (approved):

Issue #279, opened 2022-09-15
Generic Host implementation for Obsidian server


XanderStoffels:
Make Obsidian available to run as a service in a Generic Host environment. 

 - [ ] Design server runtime around a GenericHost
    - [ ] Use Console Lifetime (naturally handles CTRL+C or SIGTERM and calls IHostApplicationLifetime.StopApplication() to gracefully stop the bot as well as unblocking asynchronous runtimes). Works very well with systemd's service management on the Linux side as well.
    - [ ] Lifetime managed by IApplicationLifetime and CancellationTokens

XanderStoffels:
Updated the issue to lower the scope a bit.

PR #280, opened 2022-09-15
Generic host implementation

XanderStoffels:
Starting implementation based on https://github.com/ObsidianMC/Obsidian/issues/279.
This is a chunky one, still a work in progress.

What has been done so far:
- You can now add Obsidian as a service on a generic host.
> `services.AddObsidian()` 
- You can pass a specific server setup (`IServerSetup`) when adding Obsidian as a service. This was done because you still want to be able to load config files, worlds, etc from within the hosting application (f.e. a ConsoleApp).
> `services.AddObsidian<T>()`  where `T` is any class implementing `IServerSetup` with a parameterless constructor.
If you would like a constructor with parameters, `services.AddObsidian(T instance);` is also available.
- The server's lifetime is bound to the generic host's lifetime using cancel tokens. You can stop the generic host (and thus the server and all other services) using CTRL-C in the console.
- Logging is now setup by the startup project. Obsidian now relies on DI to get a logger.

Any feedback so far is appreciated 😄


XanderStoffels:
Closing this one because it is too big. A better way would be to split this up into multiple PR's so it is more easy to review.

PR #281, opened 2022-09-17
Obsidian now uses a Generic Host to run the Server.

### LINKED ISSUE #279: Generic Host implementation for Obsidian server

XanderStoffels:
**Hi all**,

This PR closes https://github.com/ObsidianMC/Obsidian/issues/279.

- The Console App now uses a generic host.
   - Add Obsidian to the host using `services.AddObsidian(IServerEnvironment)`, imported from `Obsidian.Hosting`.
   - Because we don't want to limit Obsidian being only able to run in a Console App, an instance of `IServerEnvironment` has to be passed. In here you can find the server configuration, saved worlds and an option that is used to indicate if the generic host should stop running if the Obsidian server stopped running. Server shutdown and crash events can also be handled here.
   - A default `IServerEnvironment` that assumes there is a file system and console input can be found in `Obsidian.Hosting`. Currently, the Console App uses this one and contains the previous logic that was already there before this PR. Can be created by calling `IServerEnvironment.CreateDefaultAsync(ILogger)`.
- Logging configuration is now done by the application that wants to run the generic host, in this case the Console App. While building the generic host, logging options are provided. The startup application is responsible for what kind of logging it wants, so `Obsidian.Logging` was moved to `Obsidian.ConsoleApp.Logging`. This makes Obsidian independent of what kind of logging the startup application wants to use (Console logging, File logging, etc) by just using Microsoft's injected `ILogger` and `ILogger<T>`.
- Fixed a small bug that would occur upon saving the world during server startup.
   - World.cs:317 ` worldFile.CopyTo($"{this.LevelDataFilePath}.old");` would crash because the file already exists. Changing it into ` worldFile.CopyTo($"{this.LevelDataFilePath}.old", true);` fixed it.
 - Partially rewrote `RconServer` to be compatible with the DI framework.

Next steps for a following pull request would be to let the Server's services being injected by the DI container. I already took some preparation for it in `Obsidian.Hosting.DependencyInjection`.

Let me know what you think. 😄


Naamloos:
I am wondering, how well does this still work with docker? I'm down to update the dockerfile and docker-compose for these updates.

for example; is it still possible to define a different working directory?

Naamloos:
never mind, considering we just call `Directory.Exists("config")` for example this should point to the current working directory. I'll finish my review in a bit.

Obsidian.API/_Interfaces/IConfig.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace Obsidian.API;
 
-public interface IConfig
+public interface IServerConfiguration
```
Naamloos, line 5: Any specific reason to rename this interface?

Obsidian.ConsoleApp/Program.cs
```diff
@@ -1,177 +1,41 @@
...
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using Obsidian.ConsoleApp.Logging;
+using Obsidian.Hosting;
+using Obsidian.Utilities;
 
 namespace Obsidian.ConsoleApp;
 
-public static class Program
+public static partial class Program
```
Naamloos: Any specific reason why you changed this to be partial? Where are the other parts?

Obsidian/Client.cs
```diff
@@ -39,7 +39,7 @@ public sealed class Client : IDisposable
     private MinecraftStream minecraftStream;
     private readonly NetworkStream socketStream;
 
-    private Config config;
+    private ServerConfiguration config;
```
Naamloos, line 42: again, any specific reason to rename? is it just for clarity? I'm down with merging this nonetheless but I just want to know the exact reasoning.

Obsidian/Hosting/DefaultServerStartup.cs
```diff
@@ -0,0 +1,116 @@
...
+        var config = new ServerConfiguration();
+
+        using var fs = configFile.Create();
+
+        await config.ToJsonAsync(fs);
+
+        Console.WriteLine($"Created new configuration file for Server");
+        Console.WriteLine($"Please fill in your config with the values you wish to use for your server.\n");
+        Console.ForegroundColor = ConsoleColor.Green;
+        Console.WriteLine(configFile.FullName);
```
Naamloos: I wonder if it's possible to update this all to use our hosting logger?

Naamloos: Just to be consistent and all.

Obsidian/Hosting/DependencyInjection.cs
```diff
@@ -0,0 +1,25 @@
+﻿using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Obsidian.Commands.Framework;
+using Obsidian.Net.Rcon;
+using Obsidian.WorldData;
+
+namespace Obsidian.Hosting;
+public static class DependencyInjection
+{
+    public static IServiceCollection AddObsidian(this IServiceCollection services, IServerEnvironment env)
```
Naamloos, line 9: now this, this is something I find beautiful. Nicely done :)

Obsidian/Server.cs
```diff
@@ -265,31 +268,31 @@ public async Task RunAsync()
...
         if (!Config.OnlineMode)
-            Logger.LogInformation($"Starting in offline mode...");
+            _logger.LogInformation($"Starting in offline mode...");
 
         Registry.RegisterCommands(this);
 
         if (Configuration.EnableRcon)
         {
-            rconServer = new RconServer(LoggerProvider.CreateLogger("RCON"), Config, this, CommandsHandler);
+            rconServer = new RconServer(_logger, Config, this, CommandsHandler);
```
Naamloos: You're adding `RconServer` as a service through `services.AddObsidian` as well... doesn't this produce two Rcon servers?

Naamloos (changes requested):
I really like the changes you've done here.

Now, I would appreciate if you'd take a look at the comments I made. It's nothing big, probably stuff that's quickly changed if needed at all.

After that I'll make sure to approve the PR.

Obsidian.API/_Interfaces/IConfig.cs
XanderStoffels: Mostly to avoid confusion. The reason I did not want it to be named something like `IConfig` or `IConfiguration` is because a generic host will automatically add a `Microsoft.Extensions.IConfiguration` to the DI container. So to avoid confusion, I made it clear this is a server configuration, not an application configuration.

Obsidian.ConsoleApp/Program.cs
XanderStoffels: Nice catch. This is a mistake on my part, I don't even remember doing that...

Obsidian/Client.cs
XanderStoffels: Yes, just for clarity. See [previous comment.](https://github.com/ObsidianMC/Obsidian/pull/281#discussion_r973659601)

Obsidian/Hosting/DefaultServerStartup.cs
XanderStoffels: It should be, yes. Let me take a look.

Obsidian/Server.cs
XanderStoffels: Nice catch! Will fix.

Obsidian/Hosting/DefaultServerStartup.cs
XanderStoffels: On second thought, this is not really Server code. You can just look at this as UI code by the startup application, so it should not be using a logger from Obsidian.

XanderStoffels:
Some feedback from @Seb-stian on from Discord has been addressed:
> Config → ServerConfiguration is nice 👍
But then using cryptic variable names (c, w, cToken, fs, env) and implicit var (where the type is nowhere to be seen) is... meh 
DefaultServerEnvironment is in a file DefaultServerStartup.cs
And its async methods don't end with Async suffix

Obsidian/Hosting/DefaultServerEnvironment.cs
```diff
@@ -0,0 +1,142 @@
...
+    Task IServerEnvironment.OnServerCrashAsync(ILogger logger, Exception e)
+    {
+        // Write crash log somewhere?
+        var byeMessages = new[]
+        {
+            "We had a good run...",
+            "At least we tried...",
+            "Who could've seen this one coming...",
+            "Try turning it off and on again...",
+            "I blame Naamloos for this one...",
```
Naamloos, line 57: kek

Naamloos (approved):

Craftplacer (approved):

PR #294, opened 2022-10-16
Rewrite KeepAlive logic to be more in line with the Notchian server.

### LINKED ISSUE #291: Rewrite keep alive logic
As it currently stands, Obsidian can handle the login and play client states, keeping the player online through the keep alive. While the keep-alive logic isn't "broke" per se, it's incorrectly implemented:

> The server will frequently send out a keep-alive, each containing a random ID. The client must respond with the same payload (see serverbound Keep Alive). If the client does not respond to them for over 30 seconds, the server kicks the client. Vice versa, if the server does not send any keep-alives for 20 seconds, the client will disconnect and yields a "Timed out" exception.

> The Notchian server uses a system-dependent time in milliseconds to generate the keep alive ID value. 

https://github.com/ObsidianMC/Obsidian/blob/c38662b4f35e979a4584a7184ab82cd1acc7a211/Obsidian/Client.cs#L468-L470

https://github.com/ObsidianMC/Obsidian/blob/c38662b4f35e979a4584a7184ab82cd1acc7a211/Obsidian/Server.cs#L663

If we take a look here at these two code examples, we can see that the method takes a `long` as [the protocol states](https://wiki.vg/Protocol#Keep_Alive_.28clientbound.29), however, it will cast the result of the subtraction to an `int`. Under perfect circumstances, this will not be an issue as the server ensures that the ids generated will always remain below `1000` and the client is intended to return the id received. However, through the use of mods, the id can be modified and return a number between `int.MaxValue` and `long.MaxValue`, which will cause an exception to be thrown. It is speculated that the server will catch the exception and handle it silently, however in the event that it does not, we can consider this to be an attack vector.

I believe the correct solution is to use Unix timestamps with millisecond precision (`DateTimeOffset.Now.ToUnixTimeMilliseconds()`) for the keep-alive id. A unix timestamp still fits in the range of `long`, prevents duplicate packet ids, and still allows us to track the player's ping.

Naamloos:
This PR would close #291.

### DIFF
--- a/Obsidian/Client.cs
+++ b/Obsidian/Client.cs
@@ -492,7 +492,7 @@ public sealed class Client : IDisposable
     {
         long keepAliveId = time.ToUnixTimeMilliseconds();
         // first, check if there's any KeepAlives that are older than 30 seconds
-        if (missedKeepAlives.Any(x => keepAliveId - x > 30000)) // magic number is 30s in millis
+        if (missedKeepAlives.Any(x => keepAliveId - x > config.KeepAliveTimeoutInterval))
         {
             // kick player, failed to respond within 30s
             cancellationSource.Cancel();
--- a/Obsidian/Client.cs
+++ b/Obsidian/Client.cs
@@ -1,4 +1,5 @@
 ﻿using Microsoft.Extensions.Logging;
+using Obsidian.API;
 using Obsidian.API.Events;
 using Obsidian.Concurrency;
 using Obsidian.Entities;
@@ -53,7 +54,7 @@ public sealed class Client : IDisposable
     private readonly Socket socket;
 
     internal int ping;
-    internal int missedKeepalives;
+    internal List<long> missedKeepAlives;
     internal int id;
 
     /// <summary>
@@ -87,6 +88,8 @@ public sealed class Client : IDisposable
 
         socketStream = new NetworkStream(socket);
         minecraftStream = new MinecraftStream(socketStream);
+
+        missedKeepAlives = new List<long>();
 
         var blockOptions = new ExecutionDataflowBlockOptions { CancellationToken = cancellationSource.Token, EnsureOrdered = true };
         packetQueue = new BufferBlock<IClientboundPacket>(blockOptions);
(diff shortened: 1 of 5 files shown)


Tides (changes requested):
We should make a config value to set the keep alive timeout and pass it down to the client so people can configure it. Beside that LGTM ~

Naamloos:
To whoever decides to do the second review: please wait before merging, I want to add the config value(s).

Obsidian/Client.cs
```diff
@@ -465,18 +468,44 @@ internal Task SendWeatherUpdateAsync()
...
+        Logger.LogDebug($"Valid KeepAlive ({keepAlive.KeepAliveId}) handled from {Player.Username} ({Player.Uuid})");
+        // KeepAlive is handled.
+        missedKeepAlives.Remove(keepAlive.KeepAliveId);
+    }
+
+    internal void SendKeepAlive(DateTimeOffset time)
+    {
+        long keepAliveId = time.ToUnixTimeMilliseconds();
+        // first, check if there's any KeepAlives that are older than 30 seconds
+        if (missedKeepAlives.Any(x => keepAliveId - x > 30000)) // magic number is 30s in millis
```
Craftplacer: small code style suggestion: use 30_000 instead of 30000, in order to make it more readable

Obsidian/Server.cs
```diff
@@ -658,12 +658,12 @@ private async Task LoopAsync()
                 await Events.InvokeServerTickAsync();
 
                 keepAliveTicks++;
-                if (keepAliveTicks > 50)
+                if (keepAliveTicks > 200) // to clarify: one tick is 50 milliseconds. 50 * 200 = 10000 millis means 10 seconds
```
Craftplacer: Make this magic number is a constant field perhaps? (comment could be moved alongside it)

Craftplacer (dismissed):

Naamloos:
Requested changes were made. Asking for re-review

Tides (approved):

Craftplacer (approved):

PR #306, opened 2022-12-27
update to 1.19.3

Tides:
Took a bit of updating but everything should be ready. Tested myself and ran into no issues


Obsidian/Net/Actions/PlayerInfo/UpdateGamemodeInfoAction.cs
```diff
@@ -5,6 +5,8 @@ public class UpdateGamemodeInfoAction : InfoAction
     public override PlayerInfoAction Type => PlayerInfoAction.UpdateGamemode;
     public int Gamemode { get; init; }
 
+    public UpdateGamemodeInfoAction(Gamemode gamemode) => this.Gamemode = Convert.ToInt32(gamemode);
```
Craftplacer: Why not type cast?

Tides: Good question I was on auto pilot 😅

Obsidian/Net/MinecraftStream.Writing.cs
```diff
@@ -345,7 +347,21 @@ public void WriteFloatAngle(Angle angle)
     public async Task WriteAngleAsync(Angle angle)
     {
         await WriteByteAsync((sbyte)angle.Value);
-        // await this.WriteUnsignedByteAsync((byte)(angle / Angle.MaxValue * byte.MaxValue));
+    }
+
+    [WriteMethod]
+    public void WriteBitSet(BitSet bitset, bool isFixed = false)
+    {
+        //TODO WE HAVE TO DO SOMETHING ABOUT THIS
```
Naamloos, line 355: What exactly is going wrong?

Naamloos (dismissed):

Obsidian/Net/MinecraftStream.Writing.cs
Tides: Right now it's assuming the length of the bitset is 1 and writing a byte but ideally we should make should sure the bitset has a inner byte array and write that instead of writing the long array if isFixed is true. It's fine for now though

PR #318, opened 2023-01-07
Block update/tags fix

Tides:
Fixes liquids not flowing and tags not adding all needed values

### DIFF
--- a/Obsidian/Net/MinecraftStream.Writing.cs
+++ b/Obsidian/Net/MinecraftStream.Writing.cs
@@ -474,11 +474,15 @@ public partial class MinecraftStream
     [WriteMethod]
     public void WriteTags(IDictionary<string, Tag[]> tagsDictionary)
     {
-        this.WriteVarInt(tagsDictionary.Count);
+        this.WriteVarInt(tagsDictionary.Count - 1);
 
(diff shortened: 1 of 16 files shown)


Obsidian/Net/MinecraftStream.Writing.cs
```diff
@@ -474,11 +474,15 @@ private void WriteTag(Tag value)
...
-        this.WriteVarInt(tagsDictionary.Count);
+        this.WriteVarInt(tagsDictionary.Count - 1);
 
         foreach (var (name, tags) in tagsDictionary)
         {
-            this.WriteString($"minecraft:{name.TrimEnd('s')}");
+            if (name == "worldgen")
+                continue;
+
+            var namespaceId = $"minecraft:{name.TrimEnd('s')}";
```
Craftplacer, line 484: `TrimEnd('s')`?

Craftplacer (approved):

Obsidian/Net/MinecraftStream.Writing.cs
Tides: In the source generators the tag categories are set after its "type" property value but mojang said I want minecraft:block not minecraft:blocks

Jonpro03 (approved):

PR #322, opened 2023-01-24
Registry refactor

Tides:
As the title suggests this is a refactor to how we handled Registry related stuff previously, using source generators it should make accessing everything much cleaner and more organized


Obsidian/Registries/Registry.cs
```diff
@@ -0,0 +1,88 @@
+﻿using Microsoft.Extensions.Logging;
+
+namespace Obsidian.Registries;
+
+public static partial class Registry
+{
+    internal static ILogger Logger { get; set; }
+
+    //public static async Task RegisterRecipesAsync()
```
Jonpro03: bruh

Jonpro03 (dismissed):
Good stuff

Obsidian/Registries/Registry.cs
Tides: Forgot to delete that class 😭

Jonpro03 (approved):

PR #329, opened 2023-03-01
Obsidian.IO

MechSlayer:
Initial implementation of the Obsidian.IO package.

Should allow for fast packet writing/reading with little to no overhead.

Right now, only the PacketWriter and some outputs are implemented.

For usage we should first have

- [ ] A clear idea about how the Writer should work.
- [ ] A clear idea about how the Reader should work.
- [ ] Tests.

After that integration could begin, we could begin by

- [ ] Update Serialization/Deserialization methods and generators to use the new package.
- [ ] Replace all usages of MinecraftStream.
- [ ] Deprecate MinecraftStream.

### DIFF
--- a/Obsidian.IO/IPacketOutput.cs
+++ b/Obsidian.IO/IPacketOutput.cs
@@ -1,0 +1,36 @@
+﻿namespace Obsidian.IO;
+
+/// <summary>
+/// Represents any type to which a packet can be written.
+/// </summary>
+public interface IPacketOutput : IDisposable
+{
+    
+    /// <summary>
+    /// The number of <see cref="byte"/> written to the output.
+    /// </summary>
+    long Written { get; }
+    
+    /// <summary>
+    /// Writes a single <see cref="byte"/> to the output.
+    /// </summary>
+    /// <param name="b"></param>
+    void WriteByte(byte b);
+    
+    /// <summary>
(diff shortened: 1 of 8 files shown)


MechSlayer:
Should the PacketReader also allow multiple input types and have more usability in exchange for performance, or make it a ref struct that only reads from a Span?

Craftplacer:
> Should the PacketReader also allow multiple input types and have more usability in exchange for performance, or make it a ref struct that only reads from a Span?

Can you provide an example how it would look like for context?

MechSlayer:
> should the PacketReader also allow multiple input types and have more usability in exchange for performance

If you look at the PacketWriter's and IPacketOutput implementations, it would be the same but for reading instead of writing.
It would allow the reader to be reusable with differenet inputs (arrays, streams, anything that can be "read" from), however this would reduce it to just a thin wrapper around whatever the input's implementation is.

> make it a ref struct that only reads from a Span

This would put the reader in charge of everything, think of it as a highly optimized Stream, but for spans, that in some cases can even inline the entire thing, however, since spans can only be stored in ref structs, and ref structs can not be stored on the heap, it would make it quite hard to use in closures, async methods, etc...

So, the first option shouldn't be too hard to integrate, however, the second one may require some heavy refactoring (mostly having to find workarounds to reading in async methods)

Jonpro03:
Abandoned

PR #331, opened 2023-03-10
Refactor UDP broadcasting of MOTD

Seb-stian:
Completed a TODO to use `PeriodicTimer` instead of `Task.Delay`. Also, MOTD gets formatted and encoded only on change, otherwise cached bytes are used.


Tides (approved):
👍🏼

Craftplacer (approved):

PR #335, opened 2023-03-16
Chunk unloading

Jonpro03:


Obsidian/WorldData/World.cs
```diff
@@ -455,6 +464,26 @@ public async Task ScheduleBlockUpdateAsync(BlockUpdate blockUpdate)
...
+
+            LoadedChunks.Except(chunksToKeep).Except(SpawnChunks).ForEach(async c =>
+            {
+                if (LoadedChunks.TryRemove(c))
+                {
+                    var r = GetRegionForChunk(c.X, c.Z);
+                    await r.UnloadChunk(c.X, c.Z);
+                }
+
+            });
```
Seb-stian: This is probably going to be really slow:
1. LINQ
2. many temporary allocations
3. file IO

Maybe add a TODO comment to optimize this? (Write out iterations instead of LINQ, avoid allocating where unnecessary, `await Task.WhenAll(...);` for IO...)

Seb-stian (dismissed):
A much needed addition.

Naamloos (dismissed):
Clean and straight to the point, I like it

Obsidian/WorldData/World.cs
Jonpro03: I dunno... what's the point of having linq if you can't use it. It reads so well.

Seb-stian (approved):

PR #336, opened 2023-03-26
Add DocFX docs

inftord:
Adds DocFX docs, and a workflow for docs to be built and published to GH pages


Craftplacer (commented):
Looks correct to me but want second opinion if this is correct.

Naamloos:
ya inf has been goin around adding docs to projects. I take it this is based on the workflow added to DSharpPlus for it's docs?

inftord:
yes, docs workflow is based on DSharpPlus's workflow for docs

Obsidian.Docs/docfx.json
```diff
@@ -0,0 +1,71 @@
+{
+  "metadata": [
+    {
+      "src": [
+        {
+          "src": "..",
+          "files": [
```
Naamloos, line 7: I think we'll only want docs for the Obsidian.API namespace. This is anything that would be available to a plugin developer. Internals are not _that_ interesting to document I think. Correct me if I'm wrong.

Naamloos (changes requested):
Other than the one comment I placed, this looks good to me.

Obsidian.Docs/docfx.json
inftord: sure, let me change that real quick

inftord: ..actually, we need to leave it like that, since im pretty sure Obsidian.API does use classes and stuff from Obsidian and Obsidian.NBT, so we need to leave it for docs to generate properly

Naamloos: I'm fairly sure it only uses NBT, not the core server.

inftord: alright then

Naamloos: nope, only references the source generators.

inftord: fixed

Naamloos (approved):

Naamloos:
Since craft already voiced their approval I'll go ahead and merge it.

inftord:
don't forget to enable GH pages in repo settings :)

PR #339, opened 2023-04-07
1.19.4 update

Tides:
Another fairly big update~ This PR made some changes to the SoundEffect part of the api if there's something you don't like about let me know. Beside that this is as usual as it is, updated packet ids and added/modified some existing one to be inline with the protocol.


Jonpro03 (approved):
I actually read through/reviewed/tested everything. I couldn't find anything worth commenting. Excellent work. I'm going to propose a 10% salary increase for you on this PR alone

Tides:
> I actually read through/reviewed/tested everything. I couldn't find anything worth commenting. Excellent work. I'm going to propose a 10% salary increase for you on this PR alone

😭

Obsidian.API/Crafting/CuttingRecipe.cs
```diff
@@ -8,7 +8,7 @@ public sealed class CuttingRecipe : IRecipe
 
     public string? Group { get; init; }
 
-    public required Ingredient Result { get; init; }
+    public required Ingredient? Result { get; init; }
```
Craftplacer, line 11: Why is this nullable?

Craftplacer (approved):

Obsidian.API/Crafting/CuttingRecipe.cs
Tides: I'll end up fixing this in another pr 👍🏼

PR #343, opened 2023-05-09
Prevent cancelled events from being un-cancelled

### LINKED ISSUE #342: Cancellable Events Can Be Uncancelled
As it stands currently, the `Cancel` property of `ICancellable` is able to be set by any plugin. This creates a possible scenario where a plugin could un-cancel an event, which may cause unexpected behavior.

The `ICancellable` interface should be replaced with an abstract class which handles the cancel logic safely:
```cs
public abstract class CancellableEvent
{
    // Gets a value indicating whether the event has been cancelled.
    public bool IsCancelled { get; private set; }

    // Sets a flag which prevents default (vanilla) behavior from executing.
    public void Cancel()
        => IsCancelled = true;
}
```

TheBrambleShark:
Fixes #342 

Overview of changes:
1. Modify `ICancellable` interface to remove setter on Cancel property.
2. Rename ICancellable.Cancel to ICancellable.IsCancelled. Hopefully this doesn't cause any issues with reflection somewhere that I'm not aware of.
3. Add ICancellable.Cancel().
4. Perform similar changes (without rename) on AsyncEventArgs so events cannot be un-handled. We can cancel this change but I felt it was good to be consistent.
5. Update consumers of ICancellable

### DIFF
--- a/Obsidian.API/Events/AsyncEventArgs.cs
+++ b/Obsidian.API/Events/AsyncEventArgs.cs
@@ -6,7 +6,15 @@
 public abstract class AsyncEventArgs : EventArgs
 {
     /// <summary>
-    /// Gets or sets whether the event was completely handled. Setting this to true will prevent remaining handlers from running.
+    /// Gets a value indicating whether the event was completely handled.
     /// </summary>
-    public bool Handled { get; set; }
+    public bool Handled { get; private set; }
+
+    /// <summary>
+    /// Marks an event as handled. Setting this flag will prevent remaining handlers from running.
+    /// </summary>
+    public void SetHandled()
+    {
+        Handled = true;
+    }
 }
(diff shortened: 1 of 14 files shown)


TheBrambleShark:
I think I got everything. Still getting a few build errors on my side, however they're items unrelated to my change (e.g. can't find a type called Material). I assume I just don't have my build environment configured properly.

Tides:
> I think I got everything. Still getting a few build errors on my side, however they're items unrelated to my change (e.g. can't find a type called Material). I assume I just don't have my build environment configured properly.

This is a source gen problem usually a clean and ide restart fixes the problem

Tides (approved):
👍🏼

Craftplacer (approved):

PR #367, opened 2023-06-06
fix: fluid interactions

Jonpro03:


Craftplacer (approved):

Obsidian/WorldData/BlockUpdates.cs
```diff
@@ -25,28 +25,55 @@ internal static async Task<bool> HandleFallingBlock(BlockUpdate blockUpdate)
...
+    /// 
+    /// </summary>
+    /// <param name="blockUpdate"></param>
+    /// <returns>Whether caller should block update neighbors</returns>
     internal static async Task<bool> HandleLiquidPhysicsAsync(BlockUpdate blockUpdate)
     {
         if (blockUpdate.Block is null) { return false; }
 
+
         var block = blockUpdate.Block;
```
Tides: You can use a little tweaked pattern matching 

```cs
if (blockUpdate.Block is not IBlock block) { return false; }
```

Tides (approved):
Looking pretty good ~

Naamloos (approved):

PR #369, opened 2023-06-07
Seal virtual object methods in PluginBase

Seb-stian:
This serves as a preparation for future plugin changes. Firstly, we ensure that behaviour of these methods is consistent for all plugins (plugins can't override it further) and secondly it improves plugin developer experience, as IDEs will only offer (via code completion) to override virtual methods that are offered to Obsidian plugins.


Tides (approved):
👍

PR #371, opened 2023-06-09
Add methods for finding entities and players in range

Seb-stian:
World now has methods to not only find entities+players in range, but also one without the other (`GetNonPlayerEntitiesInRange` and `GetPlayersInRange`). Finding entities changed behaviour - more than just the nearest region is checked now (which should be the correct behaviour).

TODO Test the implementation


Obsidian/Entities/Player.cs
Seb-stian, line 1: Why did this happen? 🤔

Tides: Maybe merge problem

Tides (dismissed):
I'll do some testing in a few here ^ ^

PR #376, opened 2023-06-10
Faster lighting

Jonpro03:
Optimize sky light spread for a new chunk. Still no x-chunk spread
Rearrange chunk building order/status to prep for x-chunk light spread

### DIFF
--- a/Obsidian/Assets/tags.json
+++ b/Obsidian/Assets/tags.json
@@ -105,7 +105,7 @@
     "name": "replaceable_by_liquid",
     "type": "blocks",
     "values": [
-      "#minecraft:replaceable_plants",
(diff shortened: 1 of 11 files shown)


Obsidian/Assets/tags.json
```diff
@@ -3219,7 +3277,7 @@
       "minecraft:flowing_lava"
     ]
   },
-  "fluids/water": {
+  "fluids/waters": {
```
Tides: Uhh This might break tags in client would recommend changing them back 😅

Tides (approved):

Obsidian/Assets/tags.json
Jonpro03: woops

PR #377, opened 2023-06-18
Core Systems Refactor

Tides:
One of the current changes is how we handled injecting/registering services from clients. ~I also want to discuss more about how we handle/register events and try to move away from scanning method names for events.~

~Draft for now as we work this out.~

As this refactor evolved, so did the systems relying on it. I did my best to keep the changes within the three big systems (PluginManager, CommandHandler and EventDispatcher).

And since I'm approaching the end of this refactor, I'd like to add the most major changes to the various systems (Events and Commands).

## PluginManager
The entire class and a majority of the things that relied on it have changed.
Here's moreso a list of things that changed.
- A service collection was created for plugins
- Each plugin container gets a scope of that service collection and is disposed when the plugin is disposed.
- Plugins have two new methods "ConfigureServices" and "ConfigureRegistry"
- All reflection methods were removed from PluginBase
- PluginDependency was removed in favor of the shared ServiceCollection
- PluginInfo is now scanned and populated from a plugin.json embedded resource. [Sample Plugin](https://github.com/ObsidianMC/Obsidian/tree/plugin-manager-cleanup/SamplePlugin)

When plugins are "handled" (e.x loaded and reloaded), two methods will be called. "ConfigureServices" and "ConfigureRegistry".
### ConfigureServices
Is, as the name explains, configuring services. These services will be added to a shared service collection and will be turned into a scope service provider for every plugin when the server is fully loaded.
### ConfigureRegistry 
Isn't as self-explanatory, at least to those unfamiliar with any type of Minecraft modding. This method is used for registering commands, events, and eventually items, blocks, entities, biomes, dimensions, and many more. And it gets called after ConfigureRegistry.

## ObjectMethodExecutor
[This Post](https://www.hanselman.com/blog/aspnet-core-architect-david-fowlers-hidden-gems-in-21) explains best why I decided to bring this in. But both EventDispatcher and CommandHandler use ObjectMethodExecutor.

## The Command Framework
The way commands are stored and executed has changed, while the system itself is more or less the same (finding and registering). Any problems we had before with commands, at least with my current testing, are not present anymore. We now support registering a "minimal API" way of registering commands as well.
```csharp
registry.MapCommand("test", 
    [CommandInfo("test command")]
    async (CommandContext ctx, int number, int otherNumber) =>
    {
        await ctx.Player.SendMessageAsync("Test #{number} and #{otherNumber}. This command was executed from the MinimalAPI.");
    });
```
Attributes and parameters are found and are read accordingly by the framework. You're not able to register command groups yet through this API, but I do plan on adding it in the future.

The way to register command modules has also changed a bit. You no longer need to provide a CommandContext parameter in your command methods as it is included as a property in the new CommandModuleBase abstract class, which all your commands should inherit from. Dependency injection works normally, either through the inject attribute or through constructors and eventually parameter injection. These classes have a scoped lifetime.

```csharp
public class MyCommands : CommandModuleBase
{
    [Inject]
    public ILogger<MyCommands> Logger { get; set; }

    [Command("mycommand")]
    [CommandInfo("woop dee doo this command is from a plugin")]
    public async Task MyCommandAsync()
    {
        Logger.LogInformation("Testing Services as injected dependency");
        await this.Player.SendMessageAsync("Hello from plugin command!");
    }
}
```

[Command Handler Internals](https://github.com/ObsidianMC/Obsidian/blob/plugin-manager-cleanup/Obsidian/Commands/Framework/CommandHandler.cs) a few things changed with this but not a lot.

## The Event System
The way events are stored and executed has *changed*, but not by a lot. Instead of scanning method names to try and match them to current events and then registering them through the traditional event subscription model, we are now checking the first parameter of the event and then adding that method to a list of executors stored in a dictionary stored by event names if it is a valid event type. The executors are then executed in ASC order of priority.

And as registering commands has a similar "minimal API" approach, so do events.

```csharp
registry.MapEvent((IncomingChatMessageEventArgs chat) =>
{
    this.Logger.LogDebug("Got a chat message! From MinimalAPI event.");
});
```

The way to register event handlers is simple. Inherit from the MinecraftEventHandler abstract class, and the PluginManager will scan and store these for you. Dependency injection works normally, either through the inject attribute or through constructors and eventually parameter injection. These classes have a scoped lifetime as well.

```csharp
public class MyEventHandler : MinecraftEventHandler
{
    [Inject]
    public ILogger<MyEventHandler> Logger { get; set; }

    [EventPriority(Priority = Priority.Critical)]
    public async ValueTask ChatEvent(IncomingChatMessageEventArgs args)
    {
        Logger.LogInformation("Wow a message through the event handler class!");
        await args.Player.SendMessageAsync("I got your chat message through the event handler class!");
    }
}
```

[Here's the internals](https://github.com/ObsidianMC/Obsidian/blob/plugin-manager-cleanup/Obsidian/Services/EventDispatcher.cs)

## Final stuff thingys
*hopefully* this will set the path for future implementations and ease up a lot of pain introduced by the old implementations.


Seb-stian:
When it comes to registering event handlers, we must have the following in mind:
- IDE support
  - discoverability via code completion
  - checking for correct signature
- support asynchronous code (handlers should return `ValueTask` or `ValueTask<>`)
- forward-compatibility - it should be possible to add events while keeping old plugins (source and binary) compatible
- enable cancellation of default behaviour

There was also a discussion about whether and how should order of handling events be determined.

Obsidian/Plugins/PluginManager.cs
```diff
@@ -226,6 +226,33 @@ public async Task UnloadPluginAsync(PluginContainer plugin)
...
+            // Registering commands from within the plugin
+            commands.RegisterCommandClass(plugin, plugin.Plugin.GetType(), plugin.Plugin);
+
+            // Registering commands found in the plugin assembly
+            var commandRoots = plugin.Plugin.GetType().Assembly.GetTypes().Where(x => x.GetCustomAttributes(false).Any(y => y.GetType() == typeof(CommandRootAttribute)));
+            foreach (var root in commandRoots)
+            {
+                commands.RegisterCommandClass(plugin, root, null);
+            }
+            CommandsRegistry.Register((Server)server);
```
Craftplacer: Separate helper method perhaps?

Craftplacer (dismissed):

Craftplacer (dismissed):

Obsidian/Plugins/PluginManager.cs
Tides: 👍🏼

Craftplacer (commented):
I'm not quite sure if using internal fields for events is a good idea, alongside assigning names to every EventArgs

Tides:
> I'm not quite sure if using internal fields for events is a good idea, alongside assigning names to every EventArgs

Yeah, it's a little pain I think I'll just get rid of the fields entirely and use the event dispatcher to register the events that way like we do with the MainCommandModule and as for the names I can just nameof the type idk why I didn't in the first place

Obsidian.API/Commands/ArgumentParsers/LocationArgumentParser.cs
```diff
@@ -1,19 +1,24 @@
...
     {
     }
 
     public override bool TryParseArgument(string input, CommandContext context, out VectorF result)
     {
         var splitted = input.Split(' ');
         var location = new VectorF();
+        var player = context.Player;
+
+        if(player is null)
```
Naamloos: Shouldn't this _only_ return when a relative position is tried to be fetched?

When the position is not relative, we don't really necessarily need the player right?

Obsidian.API/Commands/ArgumentParsers/PlayerArgumentParser.cs
```diff
@@ -0,0 +1,25 @@
...
+
+    //TODO support selectors
+    public override bool TryParseArgument(string input, CommandContext context, out IPlayer result)
+    {
+        var server = context.Server;
+        IPlayer? player = null;
+
+        if (Guid.TryParse(input, out var guid))
+            // is valid GUID, try find with guid
+            server.GetPlayer(guid);
```
Naamloos: no assignment to `player` variable- this will always return null when a valid GUID is provided.

Naamloos: With that, when for whatever crazy reason the player's username is a valid guid...

actually that's not possible I believe but might be good to handle that case anyway

Obsidian.API/Commands/CommandModuleBase.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    private CommandContext? commandContext;
+
+    [CommandContext]
+    public CommandContext CommandContext
+    {
+        get
+        {
+            if (commandContext == null)
+                throw new UnreachableException();//TODO empty command context maybe??
```
Naamloos, line 15: You can't really do anything without a context so I think just letting it throw is valid.

Obsidian.API/Utilities/CommandHelpers.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Obsidian.API.Utilities;
+internal class CommandHelpers
+{
+    public const CommandIssuers DefaultIssuerScope = CommandIssuers.Any;
+
+    public const string DefaultPrefix = "/";
```
Naamloos, line 6: Nice to have but the client always considers slash to be the default prefix

Keep it as-is though. Imagine in the future we get a different prefix, it'd be an easy swap. Doubt it though.

Obsidian.API/_Enums/Priority.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System.ComponentModel;
+
+namespace Obsidian.API;
+
+public enum Priority
```
Naamloos, line 5: Interesting as in my eyes higher priority means run first. Just an implementation detail but I am open to discuss.

Obsidian/Client.cs
```diff
@@ -315,24 +315,28 @@ public async Task StartConnectionAsync()
                 case ClientState.Configuration:
                     Debug.Assert(Player is not null);
 
-                    var configurationPacketReceived = new PacketReceivedEventArgs(Player, this.server, id, data);
+                    //await this.server.packetReceived.InvokeAsync(configurationPacketReceived);
```
Naamloos: Why is this commented out?

Obsidian/Client.cs
```diff
@@ -315,24 +315,28 @@ public async Task StartConnectionAsync()
...
 
-                    await this.server.Events.PacketReceived.InvokeAsync(configurationPacketReceived);
+                    var result = await this.server.EventDispatcher.ExecuteEventAsync(new PacketReceivedEventArgs(Player, this.server, id, data));
 
-                    if (!configurationPacketReceived.IsCancelled)
-                        await this.handler.HandleConfigurationPackets(id, data, this);
+                    if (result == EventResult.Cancelled)
+                        return;
 
+                    await this.handler.HandleConfigurationPackets(id, data, this);
```
Naamloos, line 323: oh. I see.

Obsidian/Plugins/PluginProviders/RemotePluginProvider.cs
```diff
@@ -34,13 +29,13 @@ public async Task<PluginContainer> GetPluginAsync(string path, ILogger logger)
 
         return url.Host switch
         {
-            "www.github.com" => await LoadGithubPluginAsync(path, logger, repository),
-            "gist.github.com" => await LoadGistPluginAsync(path, logger, repository),
+            "www.github.com" => await LoadGithubPluginAsync(path, repository),
+            "gist.github.com" => await LoadGistPluginAsync(path, repository),
             _ => Failed("'unknown'", path, logger, $"Remote source {url.Host} is not supported")
```
Naamloos, line 34: Might be cool to at some point allow "unsafe hosts" via a config variable

```json
{
    "ALLOW_UNSAFE_HOSTS_FOR_COMPILED_PLUGINS_LOOKOUT_DANGER_YIKES": true
}
```
idk

Obsidian/Registries/CommandsRegistry.cs
```diff
@@ -27,9 +27,9 @@ public static void Register(Server server)
                 Type = CommandNodeType.Literal
             };
 
-            foreach (var overload in cmd.Overloads.Take(1))
+            foreach (var overload in cmd.Overloads)
             {
-                var args = overload.GetParameters().Skip(1); // skipping obsidian context
+                var args = overload.GetParameters();
```
Naamloos, line 32: This may or may not break docker support, I believe this is why I added it. Might not be relevant anymore. I don't remember.

Obsidian/Services/EventDispatcher.cs
```diff
@@ -0,0 +1,210 @@
...
+using Obsidian.Events.EventArgs;
+using Obsidian.Plugins;
+using Obsidian.Utilities.Interfaces;
+using System.Collections.Frozen;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+
+namespace Obsidian.Services;
+
+//TODO BETTER NAME MAYBE??
```
Naamloos: nah

Naamloos (changes requested):
All in all, seems clean. Check my feedback, notices some minor oversights.

Obsidian.API/Commands/ArgumentParsers/LocationArgumentParser.cs
Tides: Yeah you're right 😅

Obsidian.API/Commands/ArgumentParsers/PlayerArgumentParser.cs
Tides: Will fix :+1:

Obsidian.API/_Enums/Priority.cs
Tides: Yeah, I'm open to discussing this as well. Since plugins can modify the properties of events, if the ones with higher priorities run first, then the ones with lower priorities *can* override those changes. As @Seb-stian said (he was also wondering why its sorting in ASC), give the ones with higher priority the option to cancel out the ones with lower BUT then that can cancel out the other things they could potentially be doing. And not everyone will use the priority attribute anyway, so even if we add a priority for "Listening" (meaning they aren't modifying the event in anyway and just recording changes), people might not use it and have their events thrown out for no reason. just a little tangent, we could also just use different names to make it more clear

Obsidian/Client.cs
Tides: ~We don't use this atm and plugins can't see that event either due to the IPacket interface not being exposed~ Nvm just code I forgot to delete (we use the event dispatcher now)

Obsidian/Registries/CommandsRegistry.cs
Tides: Take a look at [CommandExecutor.cs & CommandExecutor.Delegate.cs](https://github.com/ObsidianMC/Obsidian/tree/plugin-manager-cleanup/Obsidian/Commands/Framework) idk if that's relevant either but :sob:

Obsidian.API/_Enums/Priority.cs
Naamloos: the thing is that the comments in this class indicate the opposite, low goes first, high goes later :p

Obsidian/Registries/CommandsRegistry.cs
Naamloos: I'll take it.

Obsidian.API/_Enums/Priority.cs
Tides: Aye, that's intentional, as that's what's actually happening 

I'm sorry if my previous response confused you :eyes:

Naamloos (dismissed):

Naamloos:
lil mergie conflictie

Tides:
> lil mergie conflictie

I'll fix em rq

Issue #282, opened 2022-09-17
Make /time command behave like vanilla


Craftplacer:
- [ ] It does not have `set`, `add` sub-commands, but can only set a certain time (It only works like `/time set`)
- [ ] The command does not add the number of days to Day (Days through F3)
- [ ] If you write a number greater than Int32.MaxValue, the command will return an error only to the console.
- [ ]  If no time was provided, the time will be set to 1337 instead.

Qialet:
The message about 1337 was more like finding an easter egg, please leave it)

Naamloos:
Thanks for opening the issue craft <3

ghost:
World's total time is a different task, doesn't depend on the /time command. The overflow is also done separately, commands do not handle that in Obsidian as I understand, they just get values in arguments, if it's overflowing they get nothing.

PR #381, opened 2023-08-04
Make /time command behave like vanilla

ghost:
I am not sure how to implement `/time` command without arguments and `/time query` command which also takes no arguments.
What have been implemented:
- set command
- add command
 
https://github.com/ObsidianMC/Obsidian/issues/282
In the issue world's total time is a different task, doesn't depend on the /time command. The overflow is also done separately, commands do not handle that in Obsidian as I understand, they just get values in arguments, if it's overflowing they get nothing.

### DIFF
--- a/.github/ISSUE_TEMPLATE/obsidian--minecraft-server-issue.md
+++ b/.github/ISSUE_TEMPLATE/obsidian--minecraft-server-issue.md
@@ -1,0 +1,14 @@
+---
+name: 'Obsidian: Minecraft Server issue'
+about: Create a GitHub Issue for Obsidian; the minecraft server
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+Before posting your issue; please check the following box:
+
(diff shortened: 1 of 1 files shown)


abzh423:
Okay, I kinda got bored with minecraft. It sugs, will retrieve for a while.

abzh423:
If you don't need this pull requests, just close it

ghost:
Tagged: @Naamloos @Tides @Craftplacer @Seb-stian @Jonpro03 
Can someone at least give some feedback guys? I will fix the code if needed.

Tides:
Hey, I see you're using parameters instead of creating a sub command. I would recommend staying away from this until we re-implement a fixed command api/system~

ghost:
> Hey, I see you're using parameters instead of creating a sub command. I would recommend staying away from this until we re-implement a fixed command api/system~

Ok, any suggestions on what should I work?

PR #383, opened 2023-08-08
Have restructured the 'using' statements in Obsidian.ConsoleApp

ghost:
I think it makes sense to add some using statements to the global using file.


ghost:
Got more commits, will re-pull it

PR #386, opened 2023-08-13
Implemented a large amount of worldgen

explorer3039:
Added deepslate layer, bedrock layer at -64, and vanilla-like ores gen.

### DIFF
--- a/Obsidian/Assets/tags.json
+++ b/Obsidian/Assets/tags.json
@@ -98,14 +98,15 @@
       "minecraft:red_concrete_powder",
       "minecraft:black_concrete_powder",
       "minecraft:scaffolding",
-      "minecraft:dragon_egg"
+      "minecraft:dragon_egg",
+      "minecraft:gravel",
+      "minecraft:suspicious_gravel"
     ]
   },
   "blocks/replaceable_by_liquid": {
     "name": "replaceable_by_liquid",
     "type": "blocks",
     "values": [
-      "#minecraft:replaceable_by_trees",
       "#minecraft:saplings",
       "#minecraft:crops",
       "#minecraft:flowers",
@@ -113,7 +114,11 @@
       "minecraft:sweet_berry_bush",
       "minecraft:tall_grass",
       "minecraft:air",
-      "minecraft:cave_air"
+      "minecraft:cave_air",
+      "minecraft:snow",
+      "minecraft:nether_sprouts",
+      "minecraft:twisting_vines,
+      "minecraft:weeping_vines"
     ]
   },
   "banner_pattern/no_item_required": {
@@ -2075,7 +2080,6 @@
     "values": [
       "minecraft:sand",
       "minecraft:red_sand",
-      "minecraft:suspicious_sand",
       "minecraft:suspicious_sand"
     ]
   },
--- a/Obsidian/Assets/tags.json
+++ b/Obsidian/Assets/tags.json
@@ -117,7 +117,7 @@
       "minecraft:cave_air",
       "minecraft:snow",
       "minecraft:nether_sprouts",
-      "minecraft:twisting_vines,
+      "minecraft:twisting_vines",
       "minecraft:weeping_vines"
     ]
   },
--- a/Obsidian/Entities/ItemEntity.cs
+++ b/Obsidian/Entities/ItemEntity.cs
@@ -42,17 +42,17 @@ public partial class ItemEntity : Entity
         if (!CanPickup && this.TimeDropped.Subtract(DateTimeOffset.UtcNow).TotalSeconds > 1)
             this.CanPickup = true;
 
-        // foreach (var ent in this.world.GetNonPlayerEntitiesInRange(this.Position, 1.5f))
-        // {
-        //     if (ent is ItemEntity item)
-        //     {
-        //         if (item == this)
-        //             continue;
+        foreach (var ent in this.world.GetNonPlayerEntitiesInRange(this.Position, 1.5f))
+        {
+            if (ent is ItemEntity item)
+            {
+                if (item == this)
+                    continue;
 
-        //         this.Count += item.Count;
+                this.Count += item.Count;
 
-        //         await item.RemoveAsync();//TODO find a better way to removed item entities that merged
-        //     }
-        // }// remove item merging because of fix some bugs
+                await item.RemoveAsync();//TODO find a better way to removed item entities that merged
+            }
+        }
     }
 }
--- a/Obsidian/Entities/ItemEntity.cs
+++ b/Obsidian/Entities/ItemEntity.cs
@@ -39,7 +39,7 @@ public partial class ItemEntity : Entity
     {
         await base.TickAsync();
 
-        if (!CanPickup && this.TimeDropped.Subtract(DateTimeOffset.UtcNow).TotalSeconds > 5)
+        if (!CanPickup && this.TimeDropped.Subtract(DateTimeOffset.UtcNow).TotalSeconds > 1)
             this.CanPickup = true;
 
         foreach (var ent in this.world.GetNonPlayerEntitiesInRange(this.Position, 1.5f))
--- a/Obsidian/Entities/ItemEntity.cs
+++ b/Obsidian/Entities/ItemEntity.cs
@@ -39,7 +39,7 @@ public partial class ItemEntity : Entity
     {
         await base.TickAsync();
 
-        if (!CanPickup && this.TimeDropped.Subtract(DateTimeOffset.UtcNow).TotalSeconds > 1)
+        if (!CanPickup && this.TimeDropped.Subtract(DateTimeOffset.UtcNow).TotalSeconds > 5)
             this.CanPickup = true;
 
         foreach (var ent in this.world.GetNonPlayerEntitiesInRange(this.Position, 1.5f))
--- a/Obsidian/Entities/ItemEntity.cs
+++ b/Obsidian/Entities/ItemEntity.cs
@@ -42,17 +42,17 @@ public partial class ItemEntity : Entity
         if (!CanPickup && this.TimeDropped.Subtract(DateTimeOffset.UtcNow).TotalSeconds > 1)
             this.CanPickup = true;
 
-        foreach (var ent in this.world.GetNonPlayerEntitiesInRange(this.Position, 1.5f))
-        {
-            if (ent is ItemEntity item)
-            {
-                if (item == this)
-                    continue;
+        // foreach (var ent in this.world.GetNonPlayerEntitiesInRange(this.Position, 1.5f))
+        // {
+        //     if (ent is ItemEntity item)
+        //     {
+        //         if (item == this)
+        //             continue;
 
-                this.Count += item.Count;
+        //         this.Count += item.Count;
 
-                await item.RemoveAsync();//TODO find a better way to removed item entities that merged
-            }
-        }
+        //         await item.RemoveAsync();//TODO find a better way to removed item entities that merged
+        //     }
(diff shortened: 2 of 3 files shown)


Obsidian/WorldData/Generators/Overworld/ChunkBuilder.cs
```diff
@@ -109,16 +139,109 @@ internal static void Surface(GenHelper helper, Chunk chunk)
...
 
+                chunk.SetBlock(x, -64, z, BlocksRegistry.Bedrock);
                 chunk.Heightmaps[HeightmapType.WorldSurfaceWG].Set(x, z, terrainHeight);
             }
         }
     }
 
+    internal static bool GenerateOreCheck(int height, int OreType)
+    {
+        if (OreType == 0) // Coal
```
Jonpro03: Can you please rewrite the if{}else if{} as a switch statement on OreType instead?

Obsidian/WorldData/Generators/Overworld/ChunkBuilder.cs
```diff
@@ -144,9 +267,36 @@ internal static void CavesAndOres(GenHelper helper, Chunk chunk)
...
                         var oreNoise2 = helper.Noise.Ore(i + ores.Length).GetValue(worldX, y, worldZ);
-                        if (oreNoise1 > 1.0 - OreSize && oreNoise2 > 1.0 - OreSize)
+                        if (oreNoise1 > 1.0 - OreSize && oreNoise2 > 1.0 - OreSize && GenerateOreCheck(y, i))
                         {
-                            chunk.SetBlock(worldX, y, worldZ, ores[i]);
+                            if (y <= 0 && y > -64)
+                            {
+                                if (i == 6)
+                                {
+                                    if (chunk.GetBiome(x, y, z) == Biome.WindsweptHills || chunk.GetBiome(x, y, z) == Biome.WindsweptGravellyHills || chunk.GetBiome(x, y, z) == Biome.Meadow || chunk.GetBiome(x, y, z) == Biome.Grove || chunk.GetBiome(x, y, z) == Biome.SnowySlopes || chunk.GetBiome(x, y, z) == Biome.FrozenPeaks || chunk.GetBiome(x, y, z) == Biome.JaggedPeaks || chunk.GetBiome(x, y, z) == Biome.StonyPeaks)
```
Jonpro03: `chunk.GetBiome()` is a bit expensive so we should try and minimize calling it. Can you drop it into a variable so there's only one call?

Also, this can be a bit abbreviated and made easier to read with something like
```csharp
var b = chunk.GetBiome(x, y, z);
if (new List<Biome> {Biome.WindsweptHills, Biome.WindsweptGravellyHills, ...}.Contains(b))
{

}
```

Obsidian/WorldData/Generators/Overworld/ChunkBuilder.cs
```diff
@@ -144,9 +267,36 @@ internal static void CavesAndOres(GenHelper helper, Chunk chunk)
                     {
                         var oreNoise1 = helper.Noise.Ore(i).GetValue(worldX, y, worldZ);
                         var oreNoise2 = helper.Noise.Ore(i + ores.Length).GetValue(worldX, y, worldZ);
-                        if (oreNoise1 > 1.0 - OreSize && oreNoise2 > 1.0 - OreSize)
+                        if (oreNoise1 > 1.0 - OreSize && oreNoise2 > 1.0 - OreSize && GenerateOreCheck(y, i))
                         {
-                            chunk.SetBlock(worldX, y, worldZ, ores[i]);
+                            if (y <= 0 && y > -64)
```
Jonpro03: Something you might consider is to use a ternary in SetBock like so:
```csharp
chunk.SetBlock(worldX, y, worldZ, y > 0 ? ores[i] : deepores[i])
```
then you can avoid this level of if/else block

Jonpro03 (changes requested):
Thank you so much for the PR! I have a few suggestions but the only change I'd consider req'd is to change the else ifs to a switch.

Cheers!

Jonpro03 (dismissed):
Thanks for the edits! I suspect @Tides will want you to put the ItemEntity code back, even if it is broken :)
Hold tight for a second code reviewer then we'll get this merged.

Obsidian/Entities/ItemEntity.cs
```diff
@@ -39,20 +39,20 @@ public override async Task TickAsync()
...
 
-                this.Count += item.Count;
+        //         this.Count += item.Count;
 
-                await item.RemoveAsync();//TODO find a better way to removed item entities that merged
-            }
-        }
+        //         await item.RemoveAsync();//TODO find a better way to removed item entities that merged
+        //     }
+        // }// remove item merging because of fix some bugs
```
Craftplacer: This change is unrelated to the PR and hasn't been discussed or reported before. Does this fix a crash or similar? If not, I'd suggest to revert this and tackle this in another PR, preferably with a fix alongside it.

Craftplacer (changes requested):

Craftplacer (dismissed):

Obsidian/Assets/tags.json
```diff
@@ -98,22 +98,27 @@
...
+      "minecraft:dragon_egg",
+      "minecraft:gravel",
+      "minecraft:suspicious_gravel"
     ]
   },
   "blocks/replaceable_by_liquid": {
     "name": "replaceable_by_liquid",
     "type": "blocks",
     "values": [
-      "#minecraft:replaceable_by_trees",
```
Tides, line 108: Why was this removed?

Obsidian/Entities/ItemEntity.cs
```diff
@@ -39,7 +39,7 @@ public override async Task TickAsync()
     {
         await base.TickAsync();
 
-        if (!CanPickup && this.TimeDropped.Subtract(DateTimeOffset.UtcNow).TotalSeconds > 5)
+        if (!CanPickup && this.TimeDropped.Subtract(DateTimeOffset.UtcNow).TotalSeconds > 1)
```
Tides: Would revert this as the pickup delay being around 5 seconds was intentional.

Obsidian/WorldData/Generators/Overworld/ChunkBuilder.cs
```diff
@@ -109,16 +143,113 @@ internal static void Surface(GenHelper helper, Chunk chunk)
...
                     }
                 }
 
+                chunk.SetBlock(x, -64, z, BlocksRegistry.Bedrock);
                 chunk.Heightmaps[HeightmapType.WorldSurfaceWG].Set(x, z, terrainHeight);
             }
         }
     }
 
+    internal static bool GenerateOreCheck(int height, int OreType)
```
Tides: Turn OreType into an enum instead of an int so the code can be discerned easier.

Obsidian/WorldData/Generators/Overworld/ChunkBuilder.cs
```diff
@@ -144,9 +275,38 @@ internal static void CavesAndOres(GenHelper helper, Chunk chunk)
...
+                                {
+                                    chunk.SetBlock(worldX, y, worldZ, deepores[i]);
+                                }
+                            }
+                            else if (y > 0)
+                            {
+                                if (i == 6)
+                                {
+                                    var b = chunk.GetBiome(x, y, z);
+                                    if (new List<Biome> {Biome.WindsweptHills, Biome.WindsweptGravellyHills, Biome.Meadow, Biome.Grove, Biome.SnowySlopes, Biome.FrozenPeaks, Biome.JaggedPeaks, Biome.StonyPeaks}.Contains(b))
```
Tides: Instead of creating a list over and over turn this into a variable or private readonly field and call from that instead ^ ^

Obsidian/WorldData/Generators/Overworld/ChunkBuilder.cs
```diff
@@ -109,16 +143,113 @@ internal static void Surface(GenHelper helper, Chunk chunk)
...
+                break;
+            case 7:
+                if (height >= -63 && height <= 16)
+                {
+                    return true;
+                }
+                else
+                {
+                    return false;
+                }
```
Tides: Just return the condition instead of using multiple if statements ^ ^

Would also use switch pattern matching expression e.x
```cs
public bool GenerateOreCheck(int height, OreType type) => type switch
{
     OreType.WhateverOre => condition,
     OreType.WhateverOre => condition,
    ...
    _ => true
};
```
You don't need to do this ^ specifically but it would make it a bit more readable.

Tides (dismissed):
Hey! Thank you for opening up a pr to add ore to our world generation. Everything else aside from my comments looks good~

Obsidian/Assets/tags.json
explorer3039: because this contains leaves that should not be replaced by water

Obsidian/WorldData/Generators/Overworld/ChunkBuilder.cs
explorer3039: This is my first time to write c# code so I don't know how to create an enum in c#, could you help me?
Thank you so much!

Jonpro03: I can, b/c it's my awful code you're working on :)

Enums are cool b/c they basically let you give a name to a number. Instead of `6`, we get to write `OreType.Emerald`. It makes code easier to read.
I would add this enum locally:
```csharp
    internal enum OreType : int
    {
        Coal,
        Iron,
        Copper,
        Gold,
        Lapis,
        Redstone,
        Emerald,
        Diamond
    }
```

Then we get to use it. C# has a new switch syntax that makes the GenerateOreCheck function super-concise.
Instead of saying `if(condition) {return true;} else {return false;}` we can just write the condition:
```csharp
    internal static bool GenerateOreCheck(int height, OreType oreType) => oreType switch
    {
// Read this as: if the ore type is coal, return whether the height is greater than 0 and less than 320
        OreType.Coal => height >= 0 && height <= 320,
        OreType.Iron => (height >= -63 && height <= 72) || (height >= 80 && height <= 320),
        OreType.Copper => height >= -16 && height <= 112,
        OreType.Gold => height >= -63 && height <= 30,
// and so on for the rest
        _ => true
    };
```

Then later, we can loop over the enum (instead of i being an arbitrary number which we have to go find the value for)... but we have to cast it back to an `int` in places where we actually want a number, which looks like `(int)i`:
```csharp
var orePlaced = false;
foreach (OreType i in Enum.GetValues(typeof(OreType)))
{
    // Check if we should be placing a given ore at this Y level
    if (!GenerateOreCheck(y, i))
    {
        // move on to the next ore
        continue;
    }

    var b = chunk.GetBiome(x, y, z);
    // Check that Emerald is only placed in the right biomes
    if (i == OreType.Emerald && !emeraldBiomes.Contains(b))
    {
        continue;
    }

    var oreNoise1 = helper.Noise.Ore((int)i).GetValue(worldX, y, worldZ);
    var oreNoise2 = helper.Noise.Ore((int)i + ores.Length).GetValue(worldX, y, worldZ);
    if (oreNoise1 > 1.0 - OreSize && oreNoise2 > 1.0 - OreSize)
    {
        // If Y is below 0, switch to deepore varients
        chunk.SetBlock(worldX, y, worldZ, y > 0 ? ores[(int)i] : deepores[(int)i]);
        orePlaced = true;
        break;
    }
}
if (orePlaced) { continue; }
```

explorer3039: Thanks for the help!

Jonpro03 (approved):
Looks excellent. Thanks again for the PR!

Issue #96, opened 2021-04-22
Improve networking performance


Seb-stian:
Networking can be made a lot faster and more clients can be handled. The idea is to use [SocketAsyncEventArgs](https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.socketasynceventargs?view=net-5.0), pure TCP [Socket](https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.socket?view=net-5.0) and a custom lightweight buffered stream.

Resources:
https://github.com/chronoxor/NetCoreServer/blob/master/source/NetCoreServer/TcpServer.cs
https://www.codeproject.com/articles/83102/c-socketasynceventargs-high-performance-socket-cod
https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.socketasynceventargs?view=net-5.0#examples
https://github.com/ObsidianMC/Obsidian/blob/experimental/Obsidian/Net/NetWriteStream.cs

Assigning myself for now

Seb-stian:
Note to self: keep an eye on https://github.com/dotnet/runtime/issues/49941 (.NET 7?)

TheVeryStarlk:
I highly recommend checking out `Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets` as it already offers performant socket connections and listeners. It is used in ASP.NET Core so it must be 🚀

ghost:
I don't think there is much difference between TcpListener and Sockets. TcpListener uses Sockets underneath it. https://www.codeproject.com/Articles/5270779/High-Performance-TCP-Client-Server-using-TCPListen
After all the only thing we need is to accept series of bytes comming over network, the only thing we might be limited by is the serialization and deserialization of packets. Unless something in the source code was implemented incorrectly.

TheVeryStarlk:
I still think just using `Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets` is a better option, instead of going into the hassle of the needy greedy details of performant socket implementations.

PR #387, opened 2023-08-14
Migrate to use `Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets`

TheVeryStarlk:
Currently Obsidian uses the `TcpListener` for the server and raw `Socket` for the client. While those classes do the job, the issue #96 raises a concern about performance due to the fact that the currently used classes do not make use of the optimized ways for data receiving and sending. I benchmarked the login state sequence, before using the PR and after. The server averaged at 841140 ticks after the PR, and 934719 ticks before. (Results are not 100% accurate), it can also be significantly improved if we start using `System.IO.Pipelines`, and this PR is a step towards that.

### DIFF
--- a/Obsidian/Client.cs
+++ b/Obsidian/Client.cs
@@ -11,7 +11,6 @@ using Obsidian.Net.Packets.Handshaking;
 using Obsidian.Net.Packets.Login;
 using Obsidian.Net.Packets.Play;
 using Obsidian.Net.Packets.Play.Clientbound;
-using Obsidian.Net.Packets.Play.Serverbound;
 using Obsidian.Net.Packets.Status;
 using Obsidian.Registries;
 using Obsidian.Utilities.Mojang;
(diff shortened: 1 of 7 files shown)


Obsidian/Client.cs
```diff
@@ -113,7 +114,7 @@ public sealed class Client : IDisposable
     /// <summary>
     /// The socket associated with the <see cref="networkStream"/>.
     /// </summary>
-    private readonly Socket socket;
+    private readonly ConnectionContext socket;
```
Jonpro03: Let's go ahead and rename socket to connectionContext throughout

Jonpro03 (dismissed):
Looks good to me, but need more 👀 for sure

inftord (approved):
LGTM

Obsidian/Client.cs
TheVeryStarlk: Done!

Naamloos (approved):
Looks pretty good. I don't see any issues so far.

PR #388, opened 2023-08-21
 ClickContainerPacket null validation

hangardonelli:
Clicking on an empty container cell, with nothing in hand, throws a NullReferenceException

### DIFF
--- a/Obsidian/Net/Packets/Play/Serverbound/ClickContainerPacket.cs
+++ b/Obsidian/Net/Packets/Play/Serverbound/ClickContainerPacket.cs
@@ -68,9 +68,12 @@ public partial class ClickContainerPacket : IServerboundPacket
         switch (Mode)
         {
             case InventoryOperationMode.MouseClick:
-                await HandleMouseClick(container, server, player, slot);
-                break;
-
+                {
+                    if (CarriedItem == null)
+                        return;
+                    await HandleMouseClick(container, server, player, slot);
+                    break;
+                }
             case InventoryOperationMode.ShiftMouseClick:
                 {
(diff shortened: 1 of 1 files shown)


Obsidian/Net/Packets/Play/Serverbound/ClickContainerPacket.cs
```diff
@@ -68,9 +68,13 @@ public async ValueTask HandleAsync(Server server, Player player)
         switch (Mode)
         {
             case InventoryOperationMode.MouseClick:
-                await HandleMouseClick(container, server, player, slot);
-                break;
-
+                {
+                    if (CarriedItem == null)
+                        return;
```
Craftplacer, line 73: Consider using `break` instead since there's still code that would be run at the bottom.

Craftplacer (changes requested):

Obsidian/Net/Packets/Play/Serverbound/ClickContainerPacket.cs
hangardonelli: Do you mean replace `return` with `break`?
In that case, shouldn't we do it in the same way in `ShiftMouseClick` case?

Craftplacer: Looking more into how the code would behave in different situations, this seems fine actually.

Craftplacer (approved):

Jonpro03 (approved):
LGTM

Jonpro03:
Thanks for the PR!

PR #391, opened 2023-11-14
1.20.2 update

Tides:
Update for 1.20.2. Lots of changes to the client login sequence 

Any changes made in the Obsidian.Net.Packets namespace can be ignored cause its just all packet ids/a couple new packets


Obsidian/Client.cs
```diff
@@ -469,34 +500,28 @@ private async Task HandleEncryptionResponseAsync(byte[] data)
...
-        //await this.SetCompression();
-        await ConnectAsync();
+        this.SendPacket(new LoginSuccess(Player.Uuid, Player.Username)
+        {
+            SkinProperties = this.Player.SkinProperties,
+        });
     }
 
-    // TODO fix compression (.net 6)
+    // TODO fix compression now????
```
Naamloos, line 509: Do we have a GH issue open for this?

Obsidian/Client.cs
```diff
@@ -607,27 +629,15 @@ internal void SendKeepAlive(DateTimeOffset time)
             return;
         }
 
-        Logger.LogDebug($"Doing KeepAlive ({keepAliveId}) with {Player.Username} ({Player.Uuid})");
+        Logger.LogDebug("Doing KeepAlive ({keepAliveId}) with {Username} ({Uuid})", keepAliveId, Player.Username, Player.Uuid);
```
Naamloos, line 632: We should look into other uses of the logger that don't use string formatting in this way, but that's out of scope for this PR.

Obsidian/Net/Packets/ClientState.cs
```diff
@@ -5,6 +5,7 @@ public enum ClientState
     Handshaking = 0,
     Status = 1,
     Login = 2,
-    Play = 3,
+    Configuration = 3,
```
Naamloos, line 8: cool that they added a config state

Obsidian/Net/Packets/Play/Clientbound/PlayerChatMessagePacket.cs
```diff
@@ -3,21 +3,22 @@
 
 namespace Obsidian.Net.Packets.Play.Clientbound;
 
+//TODO this changed implement later
```
Naamloos, line 6: We might also want to open an issue for this to track progress.

Obsidian/Client.cs
Tides: That's a good idea we should open an issue for it

Obsidian/Net/Packets/Play/Clientbound/UpdateAdvancementsPacket.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Obsidian.Net.Packets.Play.Clientbound;
 
-//TODO finish this packet
+//TODO finish this packet PLEASE
```
Naamloos, line 6: teehee

Naamloos (approved):
Nice work as always. As stated in the previous comments we might want to open GitHub issues for the TODOs scattered through the project. This may also be useful for new contributors wanting to help us develop the server further.

PR #402, opened 2023-11-16
Bugfix/improved logging

Naamloos:


Tides (changes requested):
Giving the (parameters) a name instead of going with {0}, {1}, etc... is best practice & should be preferred ^ ^

Naamloos:
> Giving the (parameters) a name instead of going with {0}, {1}, etc... is best practice & should be preferred ^ ^

will fix that soon

Naamloos:
I think it's better to close this one for now and redo the work sometime soon.

Issue #185, opened 2021-11-14
Decouple Obsidian


Seb-stian:
Obsidian is a large project that consists of many parts. At the moment, everything sort of melts together, which is bad for maintainability and extensibility.

- [x] Separate console front-end to a new project
  - [x] Make sure that we only have things `public` when they should be
  - [x] Make types `sealed` as appropriate
- [x] Move `Obsidian.ConversionUtility` to repository `Tools`
- [x] Move `Obsidian/scripts` to repository `Tools`
- [x] Move `Obsidian.Nbt` to its own repository
- [ ] Rethink project folder/namespace structure (?)

Some tasks (?) need further discussing.

vix127:
Shouldn't `Move Obsidian.Nbt to its own repository` be checked?

PR #403, opened 2023-11-18
Server refactor

Tides:
This is a refactoring of Server, WorldManager and packet broadcasting. Abstracting them out further and removing a lot of circular dependencies.

Notable Changes:
- WorldManager is now a hosted service. It will load or generate specified worlds right after the server itself is started and also manage every loaded world chunk on its own loop. Note that this doesn't tick worlds, as that still handled within the main server loop

- Packet broadcasting is no longer handled by the server class and is now handled by the [PacketBroadcaster](https://github.com/ObsidianMC/Obsidian/blob/server-refactor/Obsidian/Services/PacketBroadcaster.cs) class & registered as a hosted service as well for priority queuing & sending packets. There are methods to skip the queue and directly send packets to clients if needed. This is the first iteration and needs more discussion.

- Entities no longer depend directly on the server to broadcast packets & will inherit the packet broadcaster service from their respective worlds. With this change, we'll have to implement more methods to do specific things (like rotating an entity's head) and not do any packet broadcasting outside of the entity class and serverbound packets.

I planned to also include the plugin manager refactor here but since its as big as these first couple changes so I left it out (#377) ~

I believe these changes are pushing the server further to be more clean and concise while tackling (#185). The end goal is to depend more on Microsoft's hosting extensions and not be overwhelmed overall.


Craftplacer (dismissed):

Jonpro03 (changes requested):
Refreshing server status during initial world gen causes a crash

Jonpro03 (commented):
Crash when loading a world from disk
EDIT: nvm... seems fine after a clean

Tides:
> Crash when loading a world from disk EDIT: nvm... seems fine after a clean

:eyes:

Naamloos:
> Crash when loading a world from disk EDIT: nvm... seems fine after a clean

Average obsidian update experience

Naamloos (dismissed):

Naamloos:
I was not aware of priorityqueue being a thing, cool!

Obsidian.API/_Interfaces/IServerConfiguration.cs
```diff
@@ -33,9 +40,13 @@ public interface IServerConfiguration
     /// Maximum amount of players that is allowed to be connected at the same time.
     /// </summary>
     public int MaxPlayers { get; set; }
+    public int PregenerateChunkRange { get; set; }
```
inftord, line 56: probably we should document all fields here, what do yall think?

Obsidian/Utilities/Extensions.cs
```diff
@@ -108,7 +105,7 @@ public static string MinecraftShaDigest(this IEnumerable<byte> data)
         // Reverse the bytes since BigInteger uses little endian
         Array.Reverse(hash);
 
-        var b = new BigInteger(hash);
+        var b = new System.Numerics.BigInteger(hash);
```
inftord, line 108: why

PR #408, opened 2023-11-24
fix a issue with cosign

inftord:
test


inftord:
technically it should fix it, but i need to check it somehow

PR #409, opened 2023-11-24
Add SetHeadRotation and SetRotation

Tides:
Adds methods to modify an entities rotation and fixes derp command :)


Naamloos:
certified poggers

Naamloos (approved):

PR #413, opened 2023-12-16
A lot of changes and a more reponsible engineer would break these out into concise PRs but I'm not that guy and write titles that are way too long

Jonpro03:


Obsidian/Client.cs
```diff
@@ -520,7 +520,6 @@ internal async Task ConnectAsync()
         Logger.LogDebug("Sent Login success to user {Username} {UUID}", Player.Username, Player.Uuid);
 
         this.State = ClientState.Play;
-        await Player.LoadAsync();
```
Tides: Why was this moved :eyes:

Jonpro03: I was trying to fix a bug that when a player logs in but they are nowhere near spawn, we fail sending them chunks and spawn them anyway (into the void). This is still broken.

But, I noticed that we get the ClientInformationPacket before Client.ConnectAsync is called. So when we call SendInfoAsync() from ClientInformationPacket, the player is not loaded yet and we send bad data.
Admittedly, LoadAsync shouldn't be in a packet handler. It was half-baked and I didn't mean to leave it in here. I quit working on that bug so I'll do this properly when I fix that bug

Tides: I already fixed that bug ^ ^

Craftplacer (dismissed):

Obsidian/Client.cs
Jonpro03: still broken, I'm afraid

Tides: Yeah I confused it for another bug but I think I know how to fix this one

Jonpro03: There is a return value on send chunks that we need to utilize

Tides (approved):

Naamloos (approved):

Naamloos:
LGTM merging

PR #417, opened 2024-01-21
Should fix #251

### LINKED ISSUE #251: Broken command [Remaining] attribute
When command is called with no arguments, command with method signature:
```cs
public void Command(CommandContext context, [Remaining] string args);
```
should work correctly with value of `args` being `""` (an empty string).

Instead, IndexOutOfRangeException is thrown. Cause can be tracked down to somewhere around [here](https://github.com/ObsidianMC/Obsidian/blob/45edc2d06136c84fcb370767a068a410a7033d6a/Obsidian/Commands/Framework/Entities/Command.cs#L114).

TheVeryStarlk:


TheVeryStarlk:
Ops, did something stupid. I will reopen this pull request.

Issue #418, opened 2024-01-21
Server crash when closing


TheVeryStarlk:
Before posting your issue; please check the following box:

- [X] I am aware that I am making a GitHub issue for the Obsidian that is a minecraft server, and not the Obsidian that is a Markdown editor.

--------------------

```
[01:16:28] [Crit]  [ObsidianHostingService] Obsidian has crashed!
[01:16:28] [Crit]  [ObsidianHostingService] Try turning it off and on again...
[01:16:28] [Crit]  [ObsidianHostingService] Reason: Object reference not set to an instance of an object.
[01:16:28] [Crit]  [ObsidianHostingService]    at Obsidian.Server.AcceptClientsAsync() in C:\Users\starl\RiderProjects\Obsidian\Obsidian\Server.cs:line 369
[01:16:28] [Crit]  [ObsidianHostingService]    at Obsidian.Server.RunAsync() in C:\Users\starl\RiderProjects\Obsidian\Obsidian\Server.cs:line 295
[01:16:28] [Crit]  [ObsidianHostingService]    at Obsidian.Server.RunAsync() in C:\Users\starl\RiderProjects\Obsidian\Obsidian\Server.cs:line 307
[01:16:28] [Crit]  [ObsidianHostingService]    at Obsidian.Hosting.ObsidianHostingService.ExecuteAsync(CancellationToken cToken) in C:\Users\starl\RiderProjects\Obsidian\Obsidian\Hosting\ObsidianHostingService.cs:line 29
```

How to re-produce:
1. Start the server
2. Go to Minecraft client and keep refreshing the server list till the server's console says: `Listening for new clients`
3. Close the server

TheVeryStarlk:
I'm trying to fix it

PR #420, opened 2024-01-21
Fixes #251 & #418

### LINKED ISSUE #251: Broken command [Remaining] attribute

TheVeryStarlk:
*(I dunno how I can separate each commit to a different PR)*

### DIFF
--- a/Obsidian/Server.cs
+++ b/Obsidian/Server.cs
@@ -67,7 +67,6 @@ public sealed partial class Server : IServer
     private readonly ILogger _logger;
 
     private IConnectionListener? _tcpListener;
-    private bool _isFullyInitialized;
 
     public ProtocolVersion Protocol => DefaultProtocol;
 
@@ -290,7 +289,6 @@ public sealed partial class Server : IServer
             continue;
 
         _logger.LogInformation("Listening for new clients...");
-        _isFullyInitialized = true;
 
         try
         {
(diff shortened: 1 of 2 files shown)


Obsidian/Server.cs
```diff
@@ -331,8 +333,17 @@ private async Task AcceptClientsAsync()
                     // No longer accepting clients.
                     break;
                 }
-
                 connection = acceptedConnection;
+
+                if (!_isFullyInitialized)
```
Tides: You can use WorldManager.ReadyToJoin

Tides (changes requested):

Naamloos:
420 blaze it

Naamloos (changes requested):
Apart from tides' requested change this looks fine

Obsidian/Server.cs
TheVeryStarlk: Should be good now 😄

Craftplacer (approved):

PR #422, opened 2024-02-09
Fix command message

TheVeryStarlk:


Craftplacer (approved):

Obsidian/Server.cs
```diff
@@ -561,6 +561,11 @@ await player.SendSoundAsync(SoundEffectBuilder.Create(SoundId.EntitySheepAmbient
             // Just stop looping.
         }
 
+        foreach (var client in _clients)
+        {
+            client.SendPacket(new DisconnectPacket(ChatMessage.Simple("Server closed"), client.State));
+        }
+
```
Craftplacer, line 568: This doesn't belong in the PR

Craftplacer (changes requested):

Obsidian/Server.cs
TheVeryStarlk: Yeah sorry. I suck at Git

TheVeryStarlk: Will re-open the PR

PR #425, opened 2024-02-16
Add lighting to island generator

Jonpro03:


Naamloos (approved):
seems reasonable enough!

PR #429, opened 2024-02-23
Fix commands not finding valid executor 

Tides:
Fixes No valid executor was found due to there being no arguments. This makes sure to use first executor if there are no arguments


Craftplacer (dismissed):
Looks alright, although a bit confused what the executor out variable is exactly doing, seems a bit repetitive/unnecessary the way it's being reassigned.

Tides:
> Looks alright, although a bit confused what the executor out variable is exactly doing, seems a bit repetitive/unnecessary the way it's being reassigned.

You're right 😅 Just fixed it up

Tides:
Okay, now commands work fully with multiple args

PR #432, opened 2024-02-27
Implement player abilities packet

TheVeryStarlk:


Obsidian/Net/Packets/Play/Clientbound/PlayerAbilitiesPacket.cs
```diff
@@ -0,0 +1,67 @@
...
+
+[Flags]
+public enum PlayerAbility
+{
+    None = 0x00,
+    Invulnerable = 0x01,
+    Flying = 0x02,
+    AllowFlying = 0x04,
+    CreativeMode = 0x08
+}
```
Tides: This should be in the API

Obsidian/Entities/Player.cs
```diff
@@ -62,7 +62,24 @@ public sealed partial class Player : Living, IPlayer
...
+            {
+                Gamemode.Creative => PlayerAbility.CreativeMode | PlayerAbility.AllowFlying | PlayerAbility.Invulnerable,
+                Gamemode.Spectator => PlayerAbility.AllowFlying | PlayerAbility.Invulnerable,
+                Gamemode.Survival or Gamemode.Adventure or Gamemode.Hardcore => PlayerAbility.None,
+                _ => throw new ArgumentOutOfRangeException(nameof(Gamemode), Gamemode, "Unknown gamemode.")
+            };
+        }
+    }
+
+    public PlayerAbility Abilities { get; internal set; }
```
Tides, line 83: This should be made accessible through the api

Obsidian/Net/Packets/Play/Clientbound/PlayerAbilitiesPacket.cs
```diff
@@ -0,0 +1,67 @@
...
+    public void Populate(MinecraftStream stream)
+    {
+        Abilities = (PlayerAbility) stream.ReadByte();
+    }
+
+    public void Populate(byte[] data)
+    {
+        using var stream = new MinecraftStream(data);
+        Populate(stream);
+    }
```
Tides, line 45: Use the Field attribute on the packet properties so the source generators can write Serialize and Populate methods for you

Tides (changes requested):

Obsidian/Net/Packets/Play/Clientbound/PlayerAbilitiesPacket.cs
TheVeryStarlk: Apparently the SG does not like it because it does not have a parameter-less constructor.

Obsidian/Net/Packets/Play/Clientbound/PlayerAbilitiesPacket.cs
TheVeryStarlk: 👍

Obsidian/Entities/Player.cs
TheVeryStarlk: 👍

Obsidian/Net/Packets/Play/Clientbound/PlayerAbilitiesPacket.cs
Tides: The Packet SG doesn't use/look at constructors 🤨

Tides: You have to make sure the class is set to partial ^ ^

TheVeryStarlk:
I will close this PR to fix the weird Git issues I am having (to not spam this PR), then re-open it again later.
cc: @Tides

PR #433, opened 2024-02-27
Implement player abilities packet

TheVeryStarlk:


Craftplacer (dismissed):
Looks fine, but doesn't seem possible to update after first packet has been sent.

TheVeryStarlk:
> Looks fine, but doesn't seem possible to update after first packet has been sent.

Yes because apparently the server should send [this instead](https://wiki.vg/Protocol#Update_Attributes)

Obsidian.API/_Types/PlayerAbility.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Obsidian.API._Types;
```
Tides: Before I merge fix pls

Tides (changes requested):
🙏🏼

Obsidian.API/_Types/PlayerAbility.cs
TheVeryStarlk: Good now?

PR #436, opened 2024-03-18
PluginManager refactor v2

Tides:
This is to remove all the old providers, as they're no longer needed, and introduce a new one that parses plugins that are packed and loads them accordingly. This did come as inspiration from the tmod format from [tModLoader](https://github.com/tModLoader/tModLoader). Something that needs more discussion is plugin signing. 

[Here's the tmod format so you don't have to look for it](https://gist.github.com/Trivaxy/a6f3550030f220894beaabc01d3bf8aa)

Right now, by default, it is not required; plugins can be packed without needing to be signed, and the server by default will also not require plugins to have a proper signature. Currently, to sign your plugin, you'll need an RSA private key written in the supported XML format needed to load them with c# (.netstandard2.0). This feature can be used differently than the way [tModLoader](https://github.com/tModLoader/tModLoader) uses them. Although I don't see a practical use for this, it does add an extra layer of complexity ~

Either way, this is tested and can be merged/improved upon. The whole goal of plugin packing is to make it easier to manage/load dependencies from plugins and to make a variety of other things easier too.


SamplePlugin/signing_key.xml
```diff
@@ -0,0 +1,16 @@
+<RSAKeyValue>
```
Naamloos, line 1: having this file be public defeats the purpose :)

Obsidian/Plugins/PluginProviders/PackedPluginProvider.cs
```diff
@@ -0,0 +1,250 @@
...
+    private readonly PluginManager pluginManager = pluginManager;
+    private readonly ILogger logger = logger;
+
+    public async Task<PluginContainer?> GetPluginAsync(string path)
+    {
+        await using var fs = new FileStream(path, FileMode.Open, FileAccess.Read, FileShare.Read);
+        using var reader = new BinaryReader(fs);
+
+        var header = Encoding.ASCII.GetString(reader.ReadBytes(4));
+        if (header != "OBBY")
```
Seb-stian: This should also work (and be more efficient)
```cs
var header = reader.ReadBytes(4);
if ("OBBY"u8.SequenceEqual(header))
```

SamplePlugin/signing_key.xml
Tides: Yeah, it's just there for testing/example 👀

Obsidian/Plugins/PluginProviders/PackedPluginProvider.cs
Tides: yessir

PR #437, opened 2024-05-04
Better Configuration

### LINKED ISSUE #407: Allow configuration through ENV variables
This would make it easier to configure Obsidian when hosting it through docker, since no files will have to be modified manually.

Tides:
Replaces the old IServerConfiguration and uses the ConfigurationManager provided by Generic Host. This also makes it possible to now use environment variables to configure the server. As long as you're familiar with how people do it in asp net Core, the same would apply here. Another PR to generally improve and make this feel more .net like 🙏🏼 closes #407


Obsidian.ConsoleApp/config/server.json
```diff
@@ -0,0 +1,30 @@
+{
```
Craftplacer, line 1: should this be committed?

Craftplacer (commented):
Aside from half of the changes you've done not matching the title of the pull request, this look fine.

Tides:
> Aside from half of the changes you've done not matching the title of the pull request, this look fine.

Lets just say they were all "necessary" changes and not me going on a minor side quest 😭 😭

Obsidian.ConsoleApp/config/server.json
Tides: yessir

Naamloos: are these embedded resources that are copied when the server first launches?

Naamloos:
looks good. Just curious about the logic of the pregenerated config files being there.

other than that, this closes #407 when merged.

Obsidian.ConsoleApp/config/server.json
Tides: They're generated when you build or publish so it'll be included in artifacts and releases

Craftplacer (approved):

PR #441, opened 2024-06-23
Expose MinecraftStream

Tides:
As the title states, this is to give developers a way to write/read packet data. An example would be through plugin messages or the new ItemComponents introduced in 1.21.

### DIFF
--- a/Obsidian.API/_Interfaces/INetSteamReader.cs
+++ b/Obsidian.API/_Interfaces/INetSteamReader.cs
@@ -1,6 +1,8 @@
 ﻿namespace Obsidian.API;
-public interface INetSteamReader
+public interface INetSteamReader : INetStream
 {
+    public bool CanRead { get; }
+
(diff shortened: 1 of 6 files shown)


Obsidian.API/_Interfaces/INetSteamReader.cs
```diff
@@ -0,0 +1,39 @@
+﻿namespace Obsidian.API;
+public interface INetSteamReader : INetStream
```
Craftplacer: Sir, we are not a games distribution platform.

Craftplacer (changes requested):
LGTM, but check my comment

Obsidian.API/_Interfaces/INetSteamReader.cs
Tides: LOL will fix 😭

Craftplacer (approved):

PR #450, opened 2024-09-13
Fix entities

Tides:
This fixes various entity issues discovered through my testing. I re-implemented the BundledDelimiterPacket (now BundledPacket) to take in a list of packets and serialize them all at once instead. That change was added for SpawnEntity as any packet alongside it and used to configure the entity needs to be processed on the same tick to avoid any issues.


Obsidian/Entities/ItemEntity.cs
```diff
@@ -35,24 +35,32 @@ public override void Write(MinecraftStream stream)
...
+
+            await item.RemoveAsync();//TODO find a better way to removed item entities that merged
         }
     }
+
+    public override async Task RemoveAsync()
+    {
+        await base.RemoveAsync();
+
+        this.world.Players.First().Value.client.server.BroadcastMessage($"Item entity removed id: {this.EntityId}");
```
Craftplacer: Should be removed later

Obsidian/WorldData/World.cs
```diff
@@ -295,19 +294,17 @@ public IEnumerable<Entity> GetNonPlayerEntitiesInRange(VectorF location, float d
...
                 // Return entities in range
                 foreach ((_, Entity entity) in region.Entities)
                 {
-                    VectorF entityLocation = entity.Position;
-                    float differenceX = entityLocation.X - location.X;
-                    float differenceY = entityLocation.Y - location.Y;
+                    var locationDifference = this.GetLocationDifference(entity.Position, location);
 
-                    if (differenceX * differenceX + differenceY * differenceY <= distance)
+                    if (locationDifference.CaluclatedDifference <= distance)
```
Craftplacer: Typo

Craftplacer (changes requested):

Obsidian/Entities/ItemEntity.cs
Tides: I forgot to remove all the logging things I did I'll do it rn

Craftplacer (approved):

PR #451, opened 2024-09-20
Update CommandHandler & Fixes the time command

### LINKED ISSUE #282: Make /time command behave like vanilla

Tides:
This updates the command handler to recognize if a type is a CommandGroup and register that alone. Also fixes the CommandRegistry from registering subcommands from a group as a root command instead of a child command.

Closes #282


Naamloos (approved):
uh lgtm

PR #453, opened 2024-11-24
Packets refactor & 2 other changes (.net9 and 1.21.3)

Tides:
The reason why this PR is beefy is due to `The symbol 'GeneratorExecutionContext' is banned for use by analyzers: Non-incremental source generators should not be used, implement IIncrementalGenerator instead`. Meaning that the PacketSerializerGenerator was broken due to not being transferred to IIncrementalSourceGenerator. I spent some time trying to figure out if I can port it over but had no luck in doing so. 

Therefore, I disabled it and opted in to serializing the packets.json we get from the data generators provided by our lovely devs @mojang in doing so that's where a majority of the changes are (re-implementing/defining all the packets that we already have). I also decided to update to the new and latest protocol version, 1.21.3. Now with packets being auto-generated, it'll make it a ton easier to implement new packets as their class names line up with the actual protocol, so it'll be simple to search for if you use MCP or something similar to decompile Minecraft. 

This will also help out a lot due to the upcoming closure of wiki.vg and potentially extremely long migration to minecraft.wiki as we'll not have much to rely on until those get updated. On the bright side, though, we can auto-generate a docs section for the packets and make our own *protocol documentation* as we start populating packets with their proper fields, etc. (Only for 1.21+ since packets.json wasn't provided in the previous versions).

This is a fairly beefy update. So I don't expect a full review and will be merging the morning after~
The only thing that's changed is how packets are generated & read/written & serialized


Naamloos:
🎉

Naamloos:
I'll give it a review either later today _or_ tomorrow after work 👍

PR #458, opened 2024-12-04
Remove consoleio

### LINKED ISSUE #284: Add a server environment for running in Docker
Currently, the default provided `IServerEnvironment` expects a terminal with standard input and output. When running in Docker, most of the time there is no standard input, causing the server to crash when waiting for input (`Console.ReadKey`, `Console.ReadLine`, etc).

Make a new IServerEnvironment that will be used by the console app based on a provided command line argument or environment variable. This environment should not expect any input from standard input.

Tides:
We don't use this for its intended purposes. and honestly we should eventually make a UI anyway.

Closes #284


Craftplacer (approved):

Craftplacer (approved):

Naamloos (approved):
LGTM. Not sure about the UI comment though. Do people actually use that?

..Could be useful for some very niche scenarios though. And debugging.

PR #459, opened 2024-12-05
1.21.4 THE CREAKING

Tides:
1.21.4 update, seems like none of the packets we use have been changed (luckily) so all it took was updating assets and serverbound ids from the handlers.

~There seems to be a bug with restarting and joining an already created world, though idk why or how this was introduced. Am trying to squash . (Actually this seems to already be present on the main branch)~

Ready for merge 🙏🏼


Naamloos (approved):
ayyy LGTM 

(MOJANG GIVE US A PACKETS.JSON THAT HAS FIELDS)

PR #460, opened 2024-12-06
Hotfix/chunk deserializing

Tides:
Fixes the problem where it would kick the user due to the biomes palette/data array not growing after deserializing


Jonpro03 (approved):
LGTM

PR #467, opened 2024-12-12
Implement ItemComponents

### LINKED ISSUE #447: De/Serialize ItemStack properly
With the update to 1.21, there have been changes made to items, causing its old format to change and being a major breaking change. Some items might not even spawn in or, in the worst case, just cause clients to disconnect or the server to throw an exception and shut down. 

[More info about the change here](https://wiki.vg/Slot_Data)

We do need to talk about how to structure this, as an ItemStack is practically a list of components now instead of the old structure. 

These are the methods/classes that should be updated.
[MinecraftStream.Writing.cs:L762-L781](https://github.com/ObsidianMC/Obsidian/blob/1.21.x/Obsidian/Net/MinecraftStream.Writing.cs#L762-L781)
[MinecraftStream.Reading.cs:L560-L670](https://github.com/ObsidianMC/Obsidian/blob/1.21.x/Obsidian/Net/MinecraftStream.Reading.cs#L560-L670)
[ItemStack.cs](https://github.com/ObsidianMC/Obsidian/blob/1.21.x/Obsidian.API/_Types/Inventory/ItemStack.cs)

Tides:
Right now we don't de/serialize ItemStacks properly as they're missing the proper metadata (components introduced in 1.20.5).  ~~I wanted to open this up to discuss/help find a good general direction on how the new API should be written. (this has led to a ton of bugs with crafting and item stack de/serializing, so it would be best if we can get this sorted out soon.~~

~~Currently, the idea that I'm working towards would involve an interface, e.g.,~~
```csharp
public interface IDataComponent
{
    public DataComponentType Type { get; }

    public string Identifier { get; }

    public void Write(INetStreamWriter writer);

    public void Read(INetStreamReader reader);
}
```

Which classes can inherit from something like:
```csharp
public sealed class MaxStackSizeItemComponent : IDataComponent
{
    public ItemComponentType Type => DataComponentType.MaxStackSize;

    public string Identifier => "minecraft:max_stack_size";

    public int MaxStackSize { get; set; }

    public MaxStackSizeItemComponent(int maxStackSize)
    {
        this.MaxStackSize = maxStackSize;
    }

    public MaxStackSizeItemComponent() 
    {
        this.MaxStackSize = 64;
    }

    public void Write(INetStreamWriter writer)
    {
        writer.WriteVarInt(this.MaxStackSize);
    }

    public void Read(INetStreamReader reader)
    {
        this.MaxStackSize = reader.ReadVarInt();
    }
}
```
And the ItemStack class would look like this, considering now it's a map of item components.
```csharp
public sealed class ItemStack : IEquatable<ItemStack>, IDictionary<DataComponentType, IDataComponent> {}
```

~~I was planning on removing ItemMeta or renaming it and making it a base class for ItemStack to inherit, but that's something we can discuss.~~ Will close #447


Obsidian.API/Inventory/DataComponents/ComponentBuilder.Map.cs
```diff
@@ -0,0 +1,77 @@
+﻿using System.Collections.Immutable;
+
+namespace Obsidian.API.Inventory.DataComponents;
+public partial class ComponentBuilder
+{
+    //TODO SOURCE GENERATE THIS
+    internal static ImmutableArray<Func<IDataComponent?>> ComponentsMap { get; } = new List<Func<IDataComponent?>>
```
Jonpro03, line 7: O_O

Jonpro03 (approved):
Yeah, that's a big boi

Obsidian.API/Inventory/DataComponents/ComponentBuilder.Map.cs
Tides: 😅

PR #474, opened 2025-01-17
Catch an exception when getting Mojang profile

TheVeryStarlk:
Also changed [this](https://github.com/TheVeryStarlk/Obsidian/blob/master/Obsidian.ConsoleApp/config/server.json#L26) to information.

### DIFF
--- a/Obsidian/Services/IUserCache.cs
+++ b/Obsidian/Services/IUserCache.cs
@@ -30,31 +30,26 @@ public sealed class UserCache(HttpClient httpClient, ILogger<UserCache> logger) 
         if (cachedUser != null && !cachedUser.Expired)
             return cachedUser;
 
-        MojangProfile? user;
-
         try
         {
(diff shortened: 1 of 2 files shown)


Obsidian/Services/IUserCache.cs
```diff
@@ -30,7 +30,17 @@ public sealed class UserCache(HttpClient httpClient, ILogger<UserCache> logger)
...
+            user = await httpClient.GetFromJsonAsync<MojangProfile>($"{userWithNameEndpoint}{escapedUsername}", Globals.JsonOptions);
+        }
+        catch
+        {
+            user = null;
+            logger.LogWarning("Could not get user profile for {Username}", username);
+        }
 
         if (user is null)
             return null;
```
Tides: Return the response in the try-catch instead of throwing it on a variable outside and just return null outside

Tides (changes requested):

Obsidian/Services/IUserCache.cs
TheVeryStarlk: Good now?

Tides (approved):

Naamloos (approved):

PR #477, opened 2025-01-30
Add ModifiedUtf8 for nbt serialization

### LINKED ISSUE #476: Nbt string serialization should use javas modified utf8 format
Right now we don't encode nbt string tag using [javas modified utf8](https://docs.oracle.com/javase/8/docs/api/java/io/DataInput.html#modified-utf-8) causing the client to immediately disconnect when sending a chat message. This extends to everything that uses nbt serialization.

More here [Nbt String Tag](https://minecraft.wiki/w/Minecraft_Wiki:Projects/wiki.vg_merge/NBT#Specification:string_tag)

Tides:
Closes #476


Obsidian.Nbt/ModifiedUtf8.cs
```diff
@@ -0,0 +1,466 @@
...
+        else
+        {
+            byteCount = GetByteCountScalar(chars);
+            return byteCount <= ushort.MaxValue;
+        }
+    }
+
+    private static int GetByteCountAvx2(ReadOnlySpan<char> chars)
+    {
+        Debug.Assert(Avx2.IsSupported);
```
Naamloos, line 408: What happens when AVX2 isn't supported? Does this lose support for e.g. raspberry pis? Have you taken this into account?

Naamloos (changes requested):
I'm not sure if I'd be willing to merge this if this means the server won't run on hardware that does not support AVX2...

Obsidian.Nbt/ModifiedUtf8.cs
```diff
@@ -0,0 +1,466 @@
...
+        }
+
+        return true;
+    }
+
+    private static void GetStringCommon(ReadOnlySpan<byte> bytes, int stringLength, ref char destination)
+    {
+        if (bytes.Length == stringLength)
+        {
+            if (!BitConverter.IsLittleEndian || stringLength < 16)
```
Naamloos, line 278: Missing a check whether Avx2 is supported

Naamloos:
Previous review can be ignored, I did see one place where a check for Avx2 support was missing. If you can add a check and make sure it properly falls back when it's not supported, it should all loog good to me and be ready to merge.

Naamloos:
Previous review can be ignored, I did see one place where a check for Avx2 support was missing. If you can add a check and make sure it properly falls back when it's not supported, it should all loog good to me and be ready to merge.

Tides:
@Naamloos can you double check make sure everything looks good before I merge ~

Naamloos:
Will do, give me a moment and I'll get to it @Tides

Naamloos (approved):

PR #478, opened 2025-01-31
Implement compressed packets

### LINKED ISSUE #393: Fix compression
As of right now Obsidian communicates uncompressed with the Minecraft client. This still has to be implemented. See the comment in the Client.cs file.

https://github.com/ObsidianMC/Obsidian/blob/4f14c58b5a564cb15fc2afeb7749044d92fad556/Obsidian/Client.cs#L430 (TODO comment)
https://github.com/ObsidianMC/Obsidian/blob/4f14c58b5a564cb15fc2afeb7749044d92fad556/Obsidian/Client.cs#L478 (SetCompression method)

RoboMico:
closes #393 

Implemented `MinecraftStream.WriteCompressedPacket` and enabled compression in `Client` class. Also rewrote `MinecraftStream.WritePacket` because how it sent bundled packets was confusing.

The code uses `System.IO.Compression.ZLibStream` built in .NET. However I couldn't find a proper scenario to test the method so all I can do is to wish [wiki.vg](https://minecraft.wiki/w/Minecraft_Wiki:Projects/wiki.vg_merge/Protocol#With_compression) was right.

### DIFF
--- a/Obsidian/Net/MinecraftStream.Writing.cs
+++ b/Obsidian/Net/MinecraftStream.Writing.cs
@@ -109,7 +109,7 @@ public partial class MinecraftStream : INetStreamWriter
         }
         else
         {   // Do not compress the packet
-            int totalLength = dataLength.GetVarIntLength() + dataLength;
+            int totalLength = 0.GetVarIntLength() + dataLength;
             dataLength = 0;
 
             this.Lock.Wait();
--- a/Obsidian/Net/MinecraftStream.Writing.cs
+++ b/Obsidian/Net/MinecraftStream.Writing.cs
@@ -97,7 +97,7 @@ public partial class MinecraftStream : INetStreamWriter
         if (dataLength >= compressionThreshold)
         {   // Compress the packet
             using MinecraftStream compressedStream = new();
-            using (ZLibStream zlibStream = new(compressedStream, CompressionLevel.Optimal))
+            using (ZLibStream zlibStream = new(compressedStream, CompressionLevel.Optimal, true))
             {
                 zlibStream.Write(dataStream.ToArray());
             }
--- a/Obsidian/Net/MinecraftStream.Writing.cs
+++ b/Obsidian/Net/MinecraftStream.Writing.cs
@@ -37,7 +37,8 @@ public partial class MinecraftStream : INetStreamWriter
     }
 
     /// <summary>
-    /// Writes a full packet to the stream.
+    /// Writes a full packet to the stream. Note that this method should NOT be an alternative to
+    /// <see cref="WriteCompressedPacket"/> for packets whose size is less than the compression threshold.
     /// </summary>
     /// <param name="packet">The packet to write.</param>
     public void WritePacket(IClientboundPacket packet)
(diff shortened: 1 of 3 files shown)


Tides (approved):
LGTM 👍🏼, I'll probably do some testing a bit later today

Naamloos (approved):
Very nice! approving for now, letting tides get to testing it before merging.

Obsidian/Net/MinecraftStream.Writing.cs
```diff
@@ -35,38 +36,91 @@ public void WriteByte(Enum value)
...
+        using MinecraftStream dataStream = new();
+        dataStream.WriteVarInt(packet.Id);
+        packet.Serialize(dataStream);
+        int dataLength = (int)dataStream.Length;
+
+        if (dataLength >= compressionThreshold)
+        {   // Compress the packet
+            using MinecraftStream compressedStream = new();
+            using ZLibStream zlibStream = new(compressedStream, CompressionLevel.Optimal);
+            zlibStream.Write(dataStream.ToArray());
```
Tides: This has to be changed to a scoped using, or the zlibstream won't flush and push the compressed data, leading to only the compression headers (2 bytes) showing up and not the entire compressed data

Obsidian/Net/MinecraftStream.Writing.cs
```diff
@@ -35,38 +36,91 @@ public void WriteByte(Enum value)
...
+        }
+        else
+        {   // Do not compress the packet
+            int totalLength = 0.GetVarIntLength() + dataLength;
+            dataLength = 0;
+
+            this.Lock.Wait();
+
+            this.WriteVarInt(totalLength);
+            this.WriteVarInt(dataLength);
```
Tides: You can just do dataLength + 1 for total length and remove datalength = 0 and just pass 0 directly instead of reassigning the var ^ ^

Tides (changes requested):
After testing quite a bit, there's a couple things I found causing problems, but besides that, everything looks good. I did notice you didn't add reading compressed packets in the client class but its fine if that's something you skipped~ Just fix up my comments and I'll merge right after ^ ^

RoboMico:
> After testing quite a bit, there's a couple things I found causing problems, but besides that, everything looks good. I did notice you didn't add reading compressed packets in the client class but its fine if that's something you skipped~ Just fix up my comments and I'll merge right after ^ ^

@Tides Thanks for reviewing! They are fixed now

Obsidian/Net/MinecraftStream.Writing.cs
```diff
@@ -35,38 +36,96 @@ public void WriteByte(Enum value)
...
+
+        using MinecraftStream dataStream = new();
+        dataStream.WriteVarInt(packet.Id);
+        packet.Serialize(dataStream);
+        int dataLength = (int)dataStream.Length;
+
+        if (dataLength >= compressionThreshold)
+        {   // Compress the packet
+            using MinecraftStream compressedStream = new();
+            using (ZLibStream zlibStream = new(compressedStream, CompressionLevel.Optimal))
```
Tides: One more thing, sorry 😅 Gotta pass `leaveOpen: true` to zlibstream so it doesn't close the backing stream as well

Tides (changes requested):

Tides (approved):

Naamloos (approved):

Naamloos:
Thank you for your contribution :)

Issue #85, opened 2021-02-07
Implement all packets


Seb-stian:
A checklist of packets that are yet to be implemented on master (ordered as on [wiki.vg](https://wiki.vg/Protocol)):

**Clientbound**

- [x] #222
- [x] Spawn Painting
- [x] Statistics
- [x] Block Entity Data
- [x] Server Difficulty
- [x] Set Cooldown
- [x] Explosion
- [ ] Tab Complete
- [x] Window Property
- [x] Disconnect
- [x] Entity Status
- [ ] Open Horse Window
- [x] Effect
- [x] #67
- [ ] Map Data
- [ ] Trade List
- [ ] Vehicle Move
- [ ] Open Book
- [ ] Open Sign Editor
- [ ] Player Abilities
- [x] Combat Event
- [x] Face Player
- [x] Player Position And Look
- [ ] Remove Entity Effect
- [ ] Resource Pack Send
- [x] Respawn
- [x] Entity Head Look
- [ ] Multi Block Change
- [ ] Select Advancement Tab
- [ ] World Border
- [ ] Camera
- [x] Held Item Change
- [ ] Update View Distance
- [ ] Attach Entity
- [x] Set Experience
- [x] Update Health
- [ ] Set Passengers
- [x] Teams
- [x] Time Update
- [x] Title
- [x] Entity Sound Effect
- [ ] Stop Sound
- [ ] NBT Query Response
- [x] Advancements
- [x] Entity Properties
- [x] Entity Effect
- [x] Declare Recipes

**Serverbound**

- [ ] Query Block NBT
- [ ] Query Entity NBT
- [ ] Set Difficulty
- [x] Client Status
- [ ] Tab Complete
- [ ] Edit Book
- [x] Interact Entity
- [ ] Generate Structure
- [ ] Lock Difficulty
- [X] Player Position And Rotation
- [X] Player Movement
- [ ] Vehicle Move
- [ ] Steer Boat
- [ ] Player Abilities
- [ ] Steer Vehicle
- [ ] Set Recipe Book State
- [ ] Resource Pack Status
- [ ] Advancement Tab
- [ ] Select Trade
- [ ] Set Beacon Effect
- [x] Held Item Change
- [ ] Update Command Block
- [ ] Update Command Block Minecart
- [ ] Update Jigsaw Puzzle
- [ ] Update Structure Block
- [ ] Update Sign
- [ ] Spectate
- [x] Use Item

**Login**
[...]

PR #480, opened 2025-02-04
Implement some clientbound packets

RoboMico:
Implement some packets in the list in #85 :

* `HorsesScreenOpenPacket`
* `MerchantOffersPacket`
* `OpenSignEditorPacket`
* `PlayerCombatEndPacket`
* `PlayerCombatEnterPacket`
* `PlayerCombatKillPacket`
* `SetChunkCacheRadiusPacket`

Note: packets are "implemented" by creating their classes and filling their fields data, the server currently does not use them. Also add some XML documents to packets related API and attribute classes.

One more thing: That list has to be updated, e.g. `OpenBookPacket` has already been implemented, while player combat event packets were not implemented until this commit.

_Fun fact: I [corrected an error](https://minecraft.wiki/?title=Minecraft_Wiki%3AProjects%2Fwiki.vg_merge%2FProtocol&diff=2845095&oldid=2844929) on the Minecraft Wiki page when checking the java source code of minecraft as the reference :)_

### DIFF
--- a/Obsidian.API/_Types/TradeEntry.cs
+++ b/Obsidian.API/_Types/TradeEntry.cs
@@ -5,7 +5,7 @@ namespace Obsidian.API;
 /// <summary>
 /// Represents a trade entry offered by a villager or wandering trader.
 /// </summary>
-public sealed record TradeEntry
+public sealed record TradeEntry : INetworkSerializable<TradeEntry>
 {
     /// <summary>
     /// The first input item of the trade. The required count of the first input is the "base price" of the trade.
@@ -28,7 +28,7 @@ public sealed record TradeEntry
     /// Whether the trade is disabled.
     /// </summary>
     public bool IsDisabled { get; set; }
-    
+
     /// <summary>
     /// Number of times the trade has been used.
     /// </summary>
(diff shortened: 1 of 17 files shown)


Obsidian.API/_Types/TradeItem.cs
```diff
@@ -0,0 +1,53 @@
+using Obsidian.API.Inventory;
+
+namespace Obsidian.API;
+
+/// <summary>
+/// Represents an item stack supplied as the "price" in a <see cref="TradeEntry"/>.
+/// </summary>
+public sealed record TradeItem
```
Tides: We have an interface `INetworkSerializable<TValue>` you can implement of off

Obsidian.API/_Types/TradeEntry.cs
```diff
@@ -0,0 +1,79 @@
+using Obsidian.API.Inventory;
+
+namespace Obsidian.API;
+
+/// <summary>
+/// Represents a trade entry offered by a villager or wandering trader.
+/// </summary>
+public sealed record TradeEntry
```
Tides: You can use INetworkSerializable<TValue> here as well ^ ^

Tides (changes requested):
Beside the 2 comments, everything looks good ~

Tides:
And yeah, you're right about #85 being a little out of date; some packet names might not match anymore either 😭

Obsidian.API/_Types/TradeItem.cs
RoboMico: `INetworkSerializable` requires a `Read` method, but I couldn't find a way to build an arbitrary data component from the net stream bytes...

`IDataComponent` does have a `Read` method: <https://github.com/ObsidianMC/Obsidian/blob/422f73c1f18d9cad2f0aa28cf5a5c8e167c653f0/Obsidian.API/Inventory/IDataComponent.cs#L10>, but it's not static so I can't create a data component object with it.

It would be best if `IDataComponent` implements `INetworkSerializable`, which I did attempt a little, but the layout of the API will be almost refactored from scratch.

Tides: > INetworkSerializable requires a Read method

Its by design not just to be used for data components; You don't need to implement the Read method, you can just `void Read() => throw new NotImplementedException()` The main use of it is to call Value.Read or Value.Write where necessary for objects serialized through the network ^ ^ Some won't need to use Read and some won't need to use Write

Tides (approved):

PR #484, opened 2025-04-24
Server refactor

### LINKED ISSUE #96: Improve networking performance

Tides:
~~Meant to open this as a draft but CHROME IS BEING SLOW....~~

This refactor includes a lot of changes to the core networking components and also attempts to decouple *most* of everything

Closes #96 

#  Core Networking
The core functionality of this has changed. Now the server and client uses [SocketAsyncEventArgs](https://learn.microsoft.com/en-us/dotnet/api/system.net.sockets.socketasynceventargs?view=net-9.0) and each client gets their own managed buffer, one for receiving and one for sending. [Server.Networking.cs](https://github.com/ObsidianMC/Obsidian/blob/server-refactor/Obsidian/Server.Networking.cs) contains almost all of the new networking stuff and the same goes for client as a majority of its core networking process has been moved to [Client.Networking.cs](https://github.com/ObsidianMC/Obsidian/blob/server-refactor/Obsidian/Client.Networking.cs). As for packet serialization that also has been changed. MinecraftStream has been removed in favor for [NetworkBuffer](https://github.com/ObsidianMC/Obsidian/blob/server-refactor/Obsidian/Net/NetworkBuffer.cs), [NetworkBuffer.Writing](https://github.com/ObsidianMC/Obsidian/blob/server-refactor/Obsidian/Net/NetworkBuffer.Writing.cs), [NetworkBuffer.Reading](https://github.com/ObsidianMC/Obsidian/blob/server-refactor/Obsidian/Net/NetworkBuffer.Reading.cs) and the same goes for its encryption counterpart. Instead of having a stream behind handling all of the reading and writing its been switched to its own internal buffer. Works *almost* like the streams api but still different. These changes should have a noticeable difference in performance when it comes to writing/reading packets and I'm sure there's things to be improved as well down the line but at least this makes it possible for upgrades to be easier in the future. A lot of the apis has been moved to be synchronous except for queuing packets that remains async. 

# Server/World/Client Decoupling
As you have notice this PR has an insane number of changes. That's due to me moving almost all of our apis away from having a direct reference to the inherited classes of World, Server and Client respectively. We shouldn't strongly depend on these and should only depend on the interfaces these classes inherit from making sure everything doesn't rely on strong internal implementations if there's something that needs to be internal and not private then the api should change into making it so that doesn't need to happen. This still isn't fully done but will be my next endeavor for the **Rework Obsidan Plan** I have came up with


Jonpro03:
0_0

Naamloos:
I am not entirely sure what you mean by the last part, "Server/World/Client Decoupling"..

Tides:
> I am not entirely sure what you mean by the last part, "Server/World/Client Decoupling"..


Decoupling May or May not be the right word to use there 😩

PR #513, opened 2026-03-12
Fix/issue 512 item dropping action

GabrielaGavi:
Updated the item drop flow so it now uses the actual stack amount the player is holding, instead of relying on a fixed value in the stack-drop action. I also adjusted DropItem() to compute a safe drop amount before spawning the entity, so the dropped item and the inventory update stay in sync.

Another important change was creating the dropped entity item from a separate ItemStack instance (with the drop amount), instead of reusing the same in-inventory object reference. This avoids side effects when the inventory count is reduced right after spawning the dropped entity.

Finally, the inventory slot update packet flow remains the same, but now it reflects the corrected drop/removal behavior more consistently from the player’s point of view.

PS: I really enjoyed your project! :)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed item dropping to properly validate that a player is holding an item before allowing drops. Items can no longer be dropped when the player's hand is empty.
  * Corrected the number of items dropped to match the actual held item count instead of using an incorrect fixed amount.

### DIFF
--- a/Obsidian/Net/Packets/Play/Serverbound/PlayerActionPacket.cs
+++ b/Obsidian/Net/Packets/Play/Serverbound/PlayerActionPacket.cs
@@ -61,9 +61,14 @@ public partial class PlayerActionPacket
                 }
             case PlayerActionStatus.DropItemStack:
                 {
-                    DropItem(player, 64);
+                    var held = player.getHeldItem();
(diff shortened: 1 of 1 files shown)


Tides:
Thank you for the PR!!.. I do have a question though, did you test the changes you made?

